PR #124, opened 2015-10-04
Changes to atom/service.py and gdata/service.py to support Domain Shared Contacts

taers232c:
Google Domain Shared Contacts requires 'GData-Version': '3.0'. These two changes support making this specification through additional_headers.
atom/service.py hardcoded 'GData-Version': '2.0'. The change only sets that value if GData-Version is not already present in additional_headers.
gdata/service.py now passes additional_headers to AtomService.__init__

### DIFF
--- a/src/atom/service.py
+++ b/src/atom/service.py
@@ -181,7 +181,8 @@ class AtomService(object):
       if content_length:
         all_headers['Content-Length'] = str(content_length)
 
-    all_headers['GData-Version'] = '2.0'
+    if 'GData-Version' not in all_headers:  
+      all_headers['GData-Version'] = '2.0'
     # Find an Authorization token for this URL if one is available.
     if self.override_token:
       auth_token = self.override_token
--- a/src/gam.py
+++ b/src/gam.py
@@ -588,7 +588,7 @@ def buildGAPIObject(api):
   try:
     service = googleapiclient.discovery.build(api, version, http=http)
   except googleapiclient.errors.UnknownApiNameOrVersion:
-    disc_filename =  u'%s-%s.json' % (api, version)
+    disc_filename = u'%s-%s.json' % (api, version)
     disc_file = os.path.join(gamSiteConfigDir, disc_filename)
     pyinstaller_disc_file = None
     try:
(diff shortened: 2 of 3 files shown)


brainysmurf:
How does passing additional_headers to its parent class make a difference, as it is just set (in the same exact way) in the child class? (GDataService is a subclass of AtomService)

taers232c:
You're right, only the atom/service.py change is required.

Thanks

brainysmurf:
Curious why are those packages being changed to accommodate? It's more portable if you override the classes. (Sorry for all the questions, I'd really like to help make gam even more awesome.)

taers232c:
If you look at the Google atom distribution, the "all_headers['GData-Version'] = '2.0'" statement is not there; it got added in somewhere along the line in Jay's version, I suppose to solve some problem where GData-Version needed to be explicitly set. My solution lets you override the value when necessary and otherwise forces the value to 2.0 as in the current version. Why should GAM have to override the class to get around a vestigial coding issue in an already customized version?

brainysmurf:
From where I'm sitting, it's better to customise packages via programable hooks (such as overriding classes), instead of copying and then customising. That means more code to maintain. But at least I understand where this is coming from now, thanks.

jay0lee:
I'm actively trying to get rid of the gdata / atom / XML libraries in GAM to simplify the codebase. XML is big fat and ugly IMHO. I know there are things it can do that JSON can't but for the Google APIs, JSON makes for a much easier dev experience.

Unfortunately it's taking Google a while to replace all of the GData APIs with Google API equivalents.

So I guess I'm ok with accepting Domain Shared Contacts GData commands into GAM as long as they aren't to invasive but I hope to see them replaced with Google API equivalents as soon as a Google API version of Contacts ships.

Regarding where to fix the GData header, Adam's right on principal but the atom / gdata libraries are already hacked up pretty well anyway. I removed all the libraries that were not GAM specific (including contacts which I assume Ross will add back in eventually).

Let's keep this patch in atom so that when atom goes away, it goes away also.

Jay

PR #134, opened 2015-11-16
Cleanup - use named constants

taers232c:
Use named constants instead of repeated literals
Define messages at top, would make language translation easier
Clean up result handling in doDownloadActivity/ExportRequest
Handle missing Python SSL module in one place

The following two files need to be updated to 3.61
https://gam-update.appspot.com/latest-version-announcement.txt
https://gam-update.appspot.com/latest-version.txt


taers232c:
Jay,

Thanks. Pull request 144 is now available.

Ross
--
ross.scroggs@gmail.com

PR #141, opened 2015-12-10
fix all user delete message iteration bug

jfalken:
When attempting to delete a message for all users, the process will
halt if we encounter a user with a mailbox that is not enabled. this
patch accounts for that case, and breaks out into the next user.

to repro; attempt to iterate through all users; where atleast one user has their mailbox disabled.

```
/gam.py all users delete messages query rfc822msgid:XXXXXXXXX@mail.gmail.com
```

if a mailbox is disabled, you will see:

```Error 400: Mail service not enabled - failedPrecondition``` 

and the script will halt.

This patch addresses this condition.

### DIFF
--- a/src/gam.py
+++ b/src/gam.py
@@ -344,6 +344,10 @@ def callGAPI(service, function, silent_errors=False, soft_errors=False, throw_re
     except googleapiclient.errors.HttpError, e:
       try:
         error = json.loads(e.content)
+        if (e.resp[u'status'] == u'400'):
+          if error['error']['message'] == u'Mail service not enabled':
+            print 'Error on current user; Mailbox not enabled. Skipping'
+            break
       except ValueError:
         if n < 3:
           disable_ssl_certificate_validation = False


jay0lee:
You can achieve the same with the code as is by running:

gam print users | gam csv - gam user ~primaryEmail delete messages query rfc822msgid:XXXXXXXXX@mail.gmail.com

then, not only will the user mailbox deletes run independent of each other, they'll also run faster as GAM will run multiple users in parallel. Default is 5 users at a time but can be bumped higher (don't recommend going higher than 20-25) by setting the GAM_THREADS environment variable.

jfalken:
I would ask that you reconsider implementing this patch. You provide a workaround, but the original bug is still present. For anyone else reading this, you can find a working fork here: https://github.com/jfalken/GAM

jay0lee:
The error is specific to Gmail API so we shouldn't try to catch it in the generic callGAPI function. I've committed a fix at:

https://github.com/jay0lee/GAM/commit/a06828dbcfdf8f78fb00e497f6bccd9825b9c33e

it tells callGAPI to handle API errors softly (just print to screen, don't exit). Let me know if this works for you. You may still want to look at the CSV method though, it really is a LOT faster.

Jay

PR #158, opened 2015-12-30
Fix typo, delete call to obsolete function

taers232c:


taers232c:
Small cleanups; Should OAuthInfo just be pass for now? #159

--
ross.scroggs@gmail.com

PR #164, opened 2016-01-01
Dynamic scopes update, other cleanup

taers232c:
I have testing scripts that, for instance, print everything that can be printed. Under the old client/service account model, this always worked fine. Now, after a few commands, I'm getting "ERROR: invalid_request: Rate limit exceeded". Are service accounts under a different set of rate restrictions?

Update. This seemed to be resolved when I went back to the API specific scope lists. I was also getting internal errors with the full scope list of 32 scopes. What won't work with the API specific scope lists?


taers232c:
See: Dynamic scope repair #165

--
ross.scroggs@gmail.com

taers232c:
Jay,

See: Dynamic scope repair/cleanup #166
This replaces #165,  #166  should apply to your latest commit.

Ross

--
ross.scroggs@gmail.com

PR #187, opened 2016-02-10
Implement gam.cfg

taers232c:
Implement configuration file gam.cfg. Documented in GamConfig.txt.

The select and config arguments allow you to manipulate gam.cfg from the command line. They are mutually exclusive and must appear before any other GAM arguments.

* select `<Section>` [save] [verify]
  * `<Section>`
    * Use this section for the current GAM command.
  * save
    * Set section = `<Section>` in the DEFAULT section
    * Write configuration data to gam.cfg
  * verify
    * Print the variable values for the current section
    * Values are determined in this order:
    * Current section
    * DEFAULT section
    * Program default

* config ((create `<Section>` [overwrite]) | (delete `<Section>`) | (select `<Section>`) |
              (make `<DestinationPath>`) | (copy `<SourceFile>` `<DestinationFile>`) |
              (reset `<VariableName>`) | (set `<VariableName>` `<Value>`) |
              save | (backup `<FileName>`) | (restore `<FileName>`) |
              verify | print)*
             [config]
  * create `<Section>` [overwrite]
    * Create this section, use it for the current GAM command
    * It is an error if `<Section>` already exists unless overwrite is specified
  * delete `<Section>`
    * Delete this section, use the DEFAULT section for the current GAM command
    * If the DEFAULT section had section = `<Section>`, clear it
  * select `<Section>`
    * Use this section for the current GAM command
  * make `<DestinationPath>`
    * `<DestinationPath>` is created
    * If `<DestinationPath>` is relative, it is appended to config_dir in DEFAULTS
  * copy `<SourceFile>` `<DestinationFile>`
    * `<SourceFile>` is copied to `<DestinationFile>`
    * `<SourceFile>` is the path of a gamscopes.json, oauth2service.json file
    * If `<SourceFile>` is relative, it is appended to GamPath
    * If `<DestinationFile>` is relative, it is appended to config_dir in the current section if defined or config_dir in DEFAULT
  * reset `<VariableName>`
    * If the current section is DEFAULT, reset the variable to its program default value
    * Else, delete the variable from the section so that the DEFAULT section value will apply
  * set `<VariableName>` `<Value>`
    * Set `<VariableName>` = `<Value>` in the current section
    * Note: section, no_browser, no_cache, no_update_check and no_verify_ssl are always set in the DEFAULT section
  * save
    * Write configuration data to gam.cfg
  * backup `<FileName>`
    * Write configuration data to `<FileName>`
    * If `<FileName>` is a relative path, it is appended to GamCfgDir
  * restore `<FileName>`
    * Read configuration data from `<FileName>`
    * If `<FileName>` is a relative path, it is appended to GamCfgDir
  * verify
    * Print the variable values for the current section
    * Values are determined in this order:
    * Current section
    * DEFAULT section
    * Program default
  * print
    * Print gam.cfg to stdout
  * config
    * Required if subsequent GAM arguments are to be processed
`<Section>` is case-sensitive except for DEFAULT which is case-insensitive.
`<VariableName>` is case-insensitive and can include the _ or not.

A new redirect argument must follow select/config if they appear but does not require that they be present. It can appear multiple times as required and must come before any other GAM commands,

* redirect csv `<FileName>` [charset `<CharSet>`]
  * If the pattern {{Section}} appears in `<FileName>`, it will be replaced with the name of the current section.
  * If `<FileName>` is relative, it is appended to drive_dir in the current section if defined or drive_dir in DEFAULT
  * If the GAM command writes a CSV file, it will be written to `<FileName>` rather than stdout
* redirect stdout `<FileName>` [append]
  * If the pattern {{Section}} appears in `<FileName>`, it will be replaced with the name of the current section.
  * If `<FileName>` is relative, it is appended to drive_dir in the current section if defined or drive_dir in DEFAULT
  * GAM output to stdout will be written to `<FileName>`
* redirect stderr `<FileName>` [append]
  * If the pattern {{Section}} appears in `<FileName>`, it will be replaced with the name of the current section.
  * If `<FileName>` is relative, it is appended to drive_dir in the current section if defined or drive_dir in DEFAULT
  * GAM output to stderr will be written to `<FileName>`

Improve support for languages with accented characters. The following commands have been extended to allow specification of the character set used in the file.
* gam batch `<FileName>` [charset `<Charset>`]
* gam csv `<FileName>` [charset `<Charset>`] gam `<GAM argument list>`

If charset is not specified, the value from gam.cfg will be used.

Allow gam csv to select rows from the CSV file.
* gam csv `<FileName>` [charset `<Charset>`] [matchfield `<FieldName>` `<PythonRegularExpression>`] gam `<GAM argument list>`

`<FieldName>` must be a value from the header row of `<FileName>`.
At its simplest, `<PythonRegularExpression>` is just text; only rows that contain that text in the `<FieldName>` column will be processed.

### DIFF
--- a/src/GamAuthorization.txt
+++ b/src/GamAuthorization.txt
@@ -1,36 +1,8 @@
+New authorization model: oauth2service.json, gamscopes.json
+
 GAM requires authorization to perform tasks on your domain; the tasks break down into two categories.
 Tasks that manipulate objects in the domain: add a user, update a group, delete a class, share a printer.
 Tasks that manipulate objects that belong to users: calendars, drive files, email.
-
-Old authorization model: oauth2service.json, client_secrets.json, oauth2.txt
-
-An OAuth Client manipulates objects in the domain; the OAuth Client acts on its own behalf to perform the tasks.
-A Service Account manipulates objects that belong to users; the Service Account acts on behalf of the user to perform the tasks.
-
-There are three broad steps to set up GAM access.
-
-1) In the Google Developers console, you specify a set of APIs that GAM will be able to use and you define an OAuth Client
-and a Service Account. This produces two files: client_secrets.json for the OAuth Client and oauth2service.json for
-the Service Account. Each of these files contain an ID that is used in the next step.
-
-2) As a domain administrator, logged into your domain Admin Console, you grant the OAuth Client ID access to one set of scopes
-and the Service Account ID access to a different set of scopes. Scopes identify an API and some or all of its functions.
-
-3) While logged into your domain Admin Console, on your computer you run the command: gam oauth create
-You select a list of scopes, GAM uses a browser to get final authorization from Google for the logged on admin user
-for these scopes and GAM writes the authorization into the file oauth2.txt that contains the email address of
-the logged on admin user and a key. You can generate multiple oauth2.txt files (with different names)
-authorizing different admins access to a different set of scopes.
-You can verify the results of this step with the command: gam oauth info
-
-This link describes the process described above: https://github.com/jay0lee/GAM/wiki/CreatingClientSecretsFile
-
-As the number of functions GAM performs has increased, a Google limit was reached. The third step above that
-produces oauth2.txt has a limit of 30 scopes which is less than the full list of scopes that GAM can use.
-You have to unselect one or two scopes to get under the limit. Currently, this is mostly annoying, but as GAMs power
-increases, it will become a bigger problem.
-
-New authorization model: oauth2service.json, gamscopes.json
 
 A Service Account manipulates objects in the domain; the Service Account acts on its own behalf to perform the tasks.
 A Service Account manipulates objects that belong to users; the Service Account acts on behalf of the user to perform the tasks.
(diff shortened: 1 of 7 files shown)


daethnir:
There's a lot here, but starting to look through it. Currently I'm working with 2 versions of gam for different security levels, based on setting the various env variables, and being able to handle this via config files would be much easier. So for me, "client" is actually "helpdesk person vs administrator".

Incindentally, it would be great if we could use files from *outside* gampath. The most annoying part of upgrading gam is getting your custom files in the newly unpacked tarball. If we could instead just point to a gam.cfg that used relative or absolute paths, then this would be a non-issue.

Since my main use case is that we'd *always* have a config and client selection, can there be an option that requires a client *always* be selected on the command line? The goal would be to avoid having defaults accidentally applied by one process that affects another. (Honestly, I'm just not a fan of having it 'remember' anything - i'd rather client needed to be always explicitly set.)


Initial impression: not sure we need an automatic cfg editor - is hand editing not sufficient?

For section name substitution, I'd think {{section}} would be better than --section-- as the former is used in various template languanges.

taers232c:
I changed --Section-- to {{Section}}

No user files are stored in gam path unless you set that to be true.

If you don't specify gamscopes_json and oauth2service_json in the Default section, then a section has to be specified with gam select for each command.

What the command line config gets you is the ability to set a variable on the fly: gam config set debug_level 4 config info user testuser1
Making sections/directories, copying config files is mostly a one time deal; it could be done in the shell and with a text editor.

PR #241, opened 2016-08-20
GAM 3.7 updates

taers232c:
Jay,

I think I've got it now.

Try gam oauth create, it should be more friendly to the user.

Ross


src/gam.py
```diff
@@ -2244,11 +2296,19 @@ def doPrintGuardians():
         states = [u'COMPLETE', u'PENDING', u'GUARDIAN_INVITATION_STATE_UNSPECIFIED']
       i += 1
     elif sys.argv[i].lower() == u'states':
-      states = sys.argv[i+1].split(u',')
+      states = sys.argv[i+1].lower().replace(u'_', u'').split(u',')
+      for j, state in enumerate(states):
+        if state not in GUARDIAN_STATE_MAP:
+          print u'ERROR: state must be one of %s; got %s' % (u', '.join(GUARDIAN_STATE_MAP), state)
```
jay0lee: The discovery doc already handles making sure state is an acceptable value. If we also check and Google adds a new state like "REJECTED" then we need to upgrade in order for GAM to use the new state. Better to let the discovery doc handle the check.

jay0lee:
This is also still not merging because you're re-adding some of my code. Might need to start clean.

taers232c:
Jay,

I finally got rid of the bear (broadcast  storm on network) that was eating my foot yesterday at one of the schools where I volunteer and now the GitHub bear is back.

I see that you've made many changes so staring clean seems like the right idea.
Do I toss my branch and re-fork; I've tried updating from your branch and that doesn't seem to get me what I want.

Thanks for your forbearance and help,

Ross

--
ross.scroggs@gmail.com

taers232c:
Jay,

I've started over. Do I issue a pull request with one change, wait for you to accept it, then do another one?

Even if I commit small changes, they pile up in the current pull request. Is there a way to make each change a separate pull request?

Ross
--
ross.scroggs@gmail.com

PR #282, opened 2016-10-12
Eliminate unused item in body

taers232c:


jay0lee:
Can you either squash the commit or sync up to master before the next PR? Seeing multiple commits here makes things messy.

PR #290, opened 2016-10-16
Use callGAPIitems to clean up code

taers232c:


src/gam.py
```diff
@@ -7899,10 +7899,10 @@ def doGetOrgInfo():
 def doGetASPs(users):
   cd = buildGAPIObject(u'directory')
   for user in users:
-    asps = callGAPI(cd.asps(), u'list', userKey=user)
-    print u'Application-Specific Passwords for %s' % user
-    try:
-      for asp in asps[u'items']:
+    asps = callGAPIitems(cd.asps(), u'list', u'items', userKey=user)
```
jay0lee: nm, i see the value in callGAPIitems at least always returning an empty list. looks good.

PR #292, opened 2016-10-16
Make set/show signature consistent.

taers232c:
Maintain backwards compatibility with Email Settings API functionality

gam user foo show signature show signature for primary email address


jay0lee:
can you clarify old Email Settings behavior vs. current source vs. your patch?

taers232c:
I think that I read too much into an example in the Migration guide to the Gmail API.

taers232c:
Do you want to go ahead and process this? Set is back to your original code and I've made show match.

PR #297, opened 2016-10-21
Version

taers232c:
Use string version rather than float.
This lets you have 3.72.0 or 3.72a; current scheme continues to work as well.

Add check to gam version  so user can determine if they're on latest version.
$ gam version check
GAM 3.72 - http://git.io/gam
Jay Lee <jay0lee@gmail.com>
Python 2.7.12 64-bit final
google-api-python-client 1.5.2
Darwin-15.6.0-x86_64-i386-64bit x86_64
Path: /Users/admin/Documents/GoogleApps/GAM
Version: Check, Current: 3.72, Latest: 3.71
$
Note that latest version at appspot is behind.

Handle bad data in lastupdatecheck.txt

### DIFF
--- a/src/gam.py
+++ b/src/gam.py
@@ -416,7 +416,7 @@ def indentMultiLineText(message, n=0):
   return message.replace(u'\n', u'\n{0}'.format(u' '*n)).rstrip()
 
 def showUsage():
-  doGAMVersion()
+  doGAMVersion(checkForCheck=False)
   print u'''
 Usage: gam [OPTIONS]...
 
--- a/src/gam.py
+++ b/src/gam.py
@@ -752,25 +752,18 @@ def SetGlobalVariables():
 
 def doGAMCheckForUpdates(forceCheck=False):
   import urllib2, calendar
-  try:
-    current_version = float(__version__)
-  except ValueError:
-    return
+  current_version = __version__
   now_time = calendar.timegm(time.gmtime())
   if not forceCheck:
-    last_check_time = readFile(GM_Globals[GM_LAST_UPDATE_CHECK_TXT], continueOnError=True, displayError=forceCheck)
-    if last_check_time == None:
-      last_check_time = 0
+    last_check_time_str = readFile(GM_Globals[GM_LAST_UPDATE_CHECK_TXT], continueOnError=True, displayError=forceCheck)
+    last_check_time = int(last_check_time_str) if last_check_time_str and last_check_time_str.isdigit() else 0
     if last_check_time > now_time-604800:
       return
   try:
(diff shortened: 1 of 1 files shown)


jay0lee:
Can we pass something like manual_check=1 to appspot query? That way I can return very latest for manual checks without needing to force update notifications for every update.

taers232c:
I'm not following. gam version check says show me my version and the latest version so I can decide if I want to got fetch an update. Otherwise, the functionality is the same as in the current version. doGAMCheckForUpdates checks lastupdate.txt and only goes to appspot if it's been a sufficiently long time since the last check.

jay0lee:
The idea is to let appspot respond to the weekly automatic check with one version but forced checks by the user with another. So when we release 3.8, I want to give it a week or so before annoying everyone with and upgrade dialog(and possible new bugs). The automatic check won't show an upgrade but the manual checks would (since admin running it seems to actually care about being on latest)

jay0lee:
nm, I'll implement it in a followup commit.

taers232c:
I see what you're saying now. So appspot would have two versions, ?manual_check=1 would get the latest and otherwise you'd get the previous version. After a week or so, you update appspot to return the save version in both cases.

jay0lee:
GitHub can do cool things:

https://developer.github.com/v3/repos/releases/

I just made a commit to switch to using these features. Now it's all automatic. I can do pre-releases and "gam version check' will show them but regular update check won't.

jay0lee:
I'm working to switch the appengine code to just point at the Github releases also but that will take a bit longer.

taers232c:
Very cool indeed, almost makes up for the GitHub bear chewing my foot occasionally.

in doPrinterRegister, change to GAL_LATEST_RELEASE
                 u'update_url': GAM_LATEST_RELEASE,

--
ross.scroggs@gmail.com

taers232c:
Suggestions:
Put a space after latest, then the two versions will line up.
     print u'Version Check\n Current: {0}\n Latest : {1}'.format(current_version, latest_version)
Put a newline after boring old version.
MESSAGE_HIT_CONTROL_C_TO_UPDATE = u'\n\nHit CTRL+C to visit the GAM website and download the latest release or wait 15 seconds continue with this boring old version.\nGAM won\'t bother you with this announcement for 1 week or you can create a file named noupdatecheck.txt in the same location as gam.py or gam.exe and GAM won\'t ever check for updates.'

--
ross.scroggs@gmail.com

taers232c:
Jay,

I've got a simple solution. Will make next PR.

Ross
--
ross.scroggs@gmail.com

PR #308, opened 2016-10-26
Handle newlines in CrOS notes field; fix print tokens to show clientId

taers232c:


jay0lee:
Did you confirm newlines show in admin console properly?  I could not get them to. API library always seems to double escape \n.

taers232c:
Jay,

$ ./gam.py update cros xxx notes 'SpareChromeBook01\nLine 1\nLine 2'
 updating xxx (1 of 1)
$ ./gam.py info  cros xxx
CrOS Device: xxx
  orgUnitPath: /Faculty
  annotatedLocation: 
  annotatedUser: 
  lastSync: 2016-10-17T06:18:34.478Z
  notes: SpareChromeBook01\nLine 1\nLine 2
  serialNumber: yyy
  status: ACTIVE
  model: Samsung Chromebook
  firmwareVersion: Google_Snow.2695.117.0
  platformVersion: 8530.96.0 (Official Build) stable-channel daisy
  osVersion: 53.0.2785.154
  bootMode: Verified
  macAddress: 5cf8a161bad6
  lastEnrollmentTime: 2015-01-09T22:41:50.065Z
  activeTimeRanges
    date: 2016-10-15
      activeTime: 8670255
      duration: 02:24:30
    date: 2016-10-16
      activeTime: 1865967
      duration: 00:31:05




Ross
--
ross.scroggs@gmail.com

jay0lee:
Did you confirm in admin console? I saw same results but admin console did not display properly.

jay0lee:
ok, seems to be working now, not sure why it didn't for me yesterday.

Regarding the other commit, I'd rather we return all items in the result rather than having a static list of which columns we'll include. That way if Google adds additional attributes to the response, GAM automatically includes them in the CSV results. Use flatten_json to deal with lists / dicts and exclude any items like kind and etag that are useless.

taers232c:
Generalized print/show tokens to not use explicit field list

PR #311, opened 2016-11-02
Fix doGamVersion

taers232c:
doGamVersion is called by showUsage with checkForArgs=False to keep
doGamVersion from wandering through some other command’s arguments

### DIFF
--- a/src/gam.py
+++ b/src/gam.py
@@ -428,7 +428,7 @@ def indentMultiLineText(message, n=0):
   return message.replace(u'\n', u'\n{0}'.format(u' '*n)).rstrip()
 
 def showUsage():
-  doGAMVersion(checkForCheck=False)
+  doGAMVersion(checkForArgs=False)
   print u'''
 Usage: gam [OPTIONS]...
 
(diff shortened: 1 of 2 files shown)


src/gam.py
```diff
@@ -818,32 +818,32 @@ def doGAMCheckForUpdates(forceCheck=False):
...
-    sys.stdout.write(__version__)
-    sys.exit(0)
-  import struct
-  version_data = u'GAM {0} - {1}\n{2}\nPython {3}.{4}.{5} {6}-bit {7}\ngoogle-api-python-client {8}\n{9} {10}\nPath: {11}'
-  print version_data.format(__version__, GAM_URL, __author__, sys.version_info[0],
-   sys.version_info[1], sys.version_info[2], struct.calcsize(u'P')*8,
-   sys.version_info[3], googleapiclient.__version__, platform.platform(),
-   platform.machine(), GM_Globals[GM_GAM_PATH])
-  if checkForCheck or force_check:
+    print __version__
```
jay0lee: print adds a newline here which we don't work. Simple is meant for (tba) upgrade shell scripts to know current version.

PR #315, opened 2016-11-04
Code fixes, cleanup

taers232c:


src/gam.py
```diff
@@ -10439,11 +10439,7 @@ def ProcessGAMCommand(args):
     elif command in [u'oauth', u'oauth2']:
       argument = sys.argv[2].lower()
       if argument in [u'request', u'create']:
-        try:
-          login_hint = sys.argv[3]
-        except IndexError:
-          login_hint = None
-        doRequestOAuth(login_hint)
+        doRequestOAuth(sys.argv[3] if len(sys.argv) > 3 else None)
```
jay0lee: Let's leave this as is, it's more lines but easier to understand what's happening.

PR #317, opened 2016-11-05
Fixes/Updates

taers232c:
GamCommands.txt
Update documentation

gam.py
doCheckServiceAccount
Correct my previous update - Lines 1211/1241

doCalendarDeleteEvent:
Eliminate duplicate events - Line 3337
Fix error message - LInes 3344/2245
Pass sendNotifications to API - LInes 3323,3349

doCalendarAddEvent
Update error message - LIne 3461

doGetCrosInfo
Add query capability - Lines 7992/8078


taers232c:
Jay,

I set it to False because the parameter is a Boolean and False is the Google default. If you believe that Google may change the default to True in the future and, therefore, None will pick up the new default, then notifyattendees should take a true/false value as you'll then have no way to turn off sendNotifications.

Ross
--
ross.scroggs@gmail.com

PR #323, opened 2016-11-09
Standardize getting login_hint for create project and oauth request

taers232c:


src/gam.py
```diff
@@ -6883,17 +6883,24 @@ def doDelProjects():
...
-    while True:
-      login_hint = raw_input(u'\nWhat is your G Suite admin email address? ')
-      if login_hint.find(u'@') == -1:
-        print u'Error: that is not a valid email address'
-      else:
-        break
+def getValidateLoginHint(login_hint):
+  def validEmailAddr(addr):
+    loc = addr.find(u'@')
+    return loc > 0 and loc < len(addr)-1
```
jay0lee: Sorry, missed this PR before pushing 4.03.

For email check, let's do something like:

http://stackoverflow.com/a/8022584/1503886

which will catch a few other bad addresses like:

user@example    # no TLD no dot
user@example@example.com  # two @ symbols

since we can't be 100% email is a valid Google account. We may want to re-prompt for the account on additional runs of create project and other commands that use this.

PR #345, opened 2016-11-19
Update runCmdForUsers

taers232c:
Eliminate default runCmdForUsers
Qualify inside runCmdForUsers whether user wants auto batch


jay0lee:
Purpose of runCmdForUsers is to allow some commands to default to batch. There's no reason IMAP updates, message deletes, etc shouldn't run in batch when it results in an order of magnitude increase in speed. Output style commands will have batch disabled by default.

We can still give users ability to choose but default should be batch for do / update commands.

PR #348, opened 2016-11-21
Set gam variables on command line; expand auto batch

taers232c:
auto_batch_min
	Automatically generate gam batch commands for commands gam `<UserTypeEntity>` ...
        If this variable is = 0, no gam batch commands will be automatically generated
        If this variable is > 0, gam batch commands will be automatically generated if the number of users in `<UserTypeEntity>` exceeds auto_batch_min
	If this variable is < 0, gam batch commands will be automatically generated for selected commands if the number of users in `<UserTypeEntity>` exceeds -auto_batch_min
        gam batch commands are never generated for gam `<UserTypeEntity>` info|print|show ...
        Default: -1, generate gam batch commands for selected commands

### DIFF
--- a/src/GamCommands.txt
+++ b/src/GamCommands.txt
@@ -18,7 +18,7 @@ Primatives
 <FalseValues>= false|off|no|disabled|0
 <DataTransferService> ::= googledrive|gdrive|drive|"drive and docs"
 <ProductID> ::= Google-Apps|Google-Coordinate|Google-Drive-storage|Google-Vault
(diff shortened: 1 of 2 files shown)


jay0lee:
Ross, I'm not going to be able to digest this large a PR at this point can you:
-hold off on the gam config stuff, I'm not comfortable adding that in until we have a simpler way to expose the config options to users (simple UI rather than a bunch more gam command arguments that aren't clear to users)
-resubmit the small fixes here like the site verification API enable as individual PRs.

taers232c:
Jay,
I put up the original PR 30 days ago and this is the first I've heard from you. I'll resubmit the small fixes this time but please provide more timely feedback in the future so I don't have to redo the work.

Ross

PR #376, opened 2016-12-31
Update licenses

taers232c:
Update documentation for new license aliases
Allow full SKU to be specified

### DIFF
--- a/src/GamCommands.txt
+++ b/src/GamCommands.txt
@@ -17,10 +17,17 @@ Primitives
 <TrueValues> ::= true|on|yes|enabled|1
 <FalseValues>= false|off|no|disabled|0
 <DataTransferService> ::= googledrive|gdrive|drive|"drive and docs"
-<ProductID> ::= Google-Apps|Google-Coordinate|Google-Drive-storage|Google-Vault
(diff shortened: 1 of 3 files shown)


src/GamCommands.txt
```diff
@@ -17,10 +17,17 @@ Primitives
...
 <FalseValues>= false|off|no|disabled|0
 <DataTransferService> ::= googledrive|gdrive|drive|"drive and docs"
-<ProductID> ::= Google-Apps|Google-Coordinate|Google-Drive-storage|Google-Vault
-<SKUID> ::= apps|gafb|gafw|gsbasic|gafg|gams|gau|unlimited|d4w|dfw|gsbusiness|coordinate|vault|vfe|
+<ProductID> ::= Google-Apps|Google-Drive-storage|Google-Vault|Google-Coordinate
+<SKUID> ::= gafb|gafw|basic|gsuite-basic|
+	    gafg|gsuite-government|
+	    gal|lite|gsuite-lite|
+	    gams|postini|gsuite-gams|
+	    gau|unlimited|gsuite-business|
```
jay0lee: Just a minor nitpick. I'd prefer to see just the "canonical" command listed instead of every possible variation which does the exact same thing. For example, just list "gsuite-business" as the SKU instead of all the other possible strings that do the same thing. I leave the multiple options around mostly for backwards compatability and to try and do things right if user enters things a slightly different way.

PR #377, opened 2016-12-31
When default SKU/Product list is used, sort it so output is cleaner

taers232c:


jay0lee:
Can you make this a separate PR?

PR #389, opened 2017-01-15
Allow field selection/deselection in gam print courses

taers232c:


jay0lee:
I'd like to see these reworked so that they dynamically determine possible fields. That way when Google provides additional fields on their end GAM simply supports it w/o a code change.

PR #406, opened 2017-01-26
gam group <GroupName> should only return users

taers232c:
When issuing a command like: `gam group <GroupName> command ...`
You want only the members of the group of type USER returned, otherwise, the command will throw an exception when a member of type GROUP or CUSTOMER is encountered and no further group members will be processed.

It is especially painful in this case: `gam course <CourseID> sync students group <GroupName>` if
`<GroupName>` contains a member of type GROUP. Why would it contain a member of type GROUP? Who knows but if you add a group email address to a course, you're going to be very unhappy.

Of course, when you say: `gam update group <GroupName> add/delete/sync members group <AnotherGroupName>`, you do want all of the members of `<AnotherGroupName>` regardless of type.

With this change, users will notice less exception throwing, a good thing.

The other noticeable change is here: `gam group <GroupName> print`
This will only print the members of type USER.
This gets the entire membership `gam print group-members group <GroupName>`

### DIFF
--- a/src/gam.py
+++ b/src/gam.py
@@ -6575,7 +6575,7 @@ def doUpdateGroup():
         checkNotSuspended = True
         i += 1
       if sys.argv[i].lower() in usergroup_types:
-        users_email = getUsersToModify(entity_type=sys.argv[i], entity=sys.argv[i+1], checkNotSuspended=checkNotSuspended)
+        users_email = getUsersToModify(entity_type=sys.argv[i], entity=sys.argv[i+1], checkNotSuspended=checkNotSuspended, groupUserMembersOnly=False)
       else:
         users_email = [sys.argv[i],]
       for user_email in users_email:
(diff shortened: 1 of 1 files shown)


jay0lee:
* We can probably kill "gam group <email> print" and similar commands. It was a hack from a long time ago and it'd be better for users to use correct commands.
* Might make sense to add something like "gam extendedgroup <email> ..." that gets all domain users if CUSTOMER is in group and gets all members of groups that are a member of the group (inheritance) but agree this is good for now to solve current errors.

runner7532:
Using the command `gam group <groupname>` it returns the following:

> Getting all members of it@company.com (may take some time for large groups)...
Got 11 all members...
GAM 4.12 - http://git.io/gam
Jay Lee <jay0lee@gmail.com>
Python 2.7.12 64-bit final
google-api-python-client 1.6.1
Windows-10-10.0.14393 AMD64
Path: C:\GAM

>Usage: gam [OPTIONS]...

>GAM. Retrieve or set G Suite domain,
user, group and alias settings. Exhaustive list of commands
can be found at: https://github.com/jay0lee/GAM/wiki

>Examples:
gam info domain
gam create user jsmith firstname John lastname Smith password secretpass
gam update user jsmith suspended on
gam.exe update group announcements add member jsmith
...

It is not actually returning the members of the group.

taers232c:
runner7532,

If you're trying to list the email addresses of the members of the group, say: gam group <groupname> print
Your command doesn't specify an action: gam group <groupname>

Ross
--
ross.scroggs@gmail.com

PR #409, opened 2017-01-28
Optimize doProcessMessagesOrThreads

taers232c:


jay0lee:
I'm not seeing significant advantages to these changes. A few variables do need to be cleaned up to be unit instead of message but otherwise I'm not seeing needs for change.

taers232c:
If you don't see a significant advantage, toss it, it's your code.
--
ross.scroggs@gmail.com

taers232c:
At a minimum make this change:
    listResult = callGAPIpages(unitmethod(), u'list', unit, page_message=page_message,
                               userId=u'me', q=query, includeSpamTrash=True, soft_errors=True, fields=u'nextPageToken,{0}(id)'.format(unit))
When fields is not specified, you get id and threadId for messages and id, snippet and historyId for threads.
All you want is id, all the other data is wasted bandwidth.

--
ross.scroggs@gmail.com

PR #411, opened 2017-01-31
Handle customer id in group add/delete/sync/clear 

taers232c:


jay0lee:
Historically it was possibe to add * to the group in order to add "all users in domain". We should make sure * is still interpreted correctly by GAM to add all users.

taers232c:
It would appear to me that * never could have worked in managing groups; here's the old code from the bottom of  getUsersToModify; * would have been replaced by *domain.com so the test in updateGroups would have never seen a plain *.
```   for user in users:
        if user[:4] == u'uid:':
          full_users.append(user[4:])
       elif user.find(u'@') == -1:
          full_users.append(u'%s@%s' % (user, GC_Values[GC_DOMAIN]))
        else:
          full_users.append(user)
```

PR #414, opened 2017-02-03
early support to get info user info about 2-step verification enrollment

hajdbo:
Google now gives information about the 2SV status of users in Admin API.
https://developers.google.com/admin-sdk/directory/v1/reference/users#resource

The previous workaround was to use Reports API, which often lagged days to report a change.


jay0lee:
Merged, thanks for the PR!

hajdbo:
another coming for "print users"

PR #420, opened 2017-02-04
Initial cleanup of gam resoldcustomer/resoldsubscription

taers232c:


jay0lee:
I still have some big work to do on this function so let's hold on this for now.

taers232c:
I changed customer_id to customerId to make the variable match the API argument as per your request on the smime code yesterday.
I cleaned up the error messages, and handled cases where a mistyped command line argument caused an infinite loop.
The rest of the changes are bug fixes.
If you apply this PR now, I won't have to redo the work in the future and you're working from a better base.

Ross

--
ross.scroggs@gmail.com

jay0lee:
It was to late to merge this since I'd already worked on adding plan change functionality and other changes but I incorporated most of your changes in my recent commit. Regarding customer_id vs customerId, I broke from the preference because I was concerned a customerId variable would conflict with the global setting (e.g. my_customer) but that doesn't seem to be the case.

taers232c:
I'll watch for that commit.

PR #432, opened 2017-02-12
Clean up SKU handling

taers232c:
As the display name of SKUs is being written to CSV files, it has to be accepted on input

In print users, the skuIds values need to be separated by something other that space as the display names have spaces in them

I'm wondering if writing the display names to the CSV files is going to break existing users scripts. In print licenses, we could leave the skuId column with the actual SKU ID and add a new column skuDisplay with the SKU display name.
In print users, we could leave the skuId column with space separated actual SKU IDs and add a new column skuDisplay with comma (or some other character) separated SKU display names

In the info user Licenses section, we could show SKU ID (SKU display name)

### DIFF
--- a/src/gam.py
+++ b/src/gam.py
@@ -4351,18 +4351,14 @@ def getImap(users):
         print u'User: {0}, IMAP Enabled: {1} ({2}/{3})'.format(user, enabled, i, count)
 
 def getProductAndSKU(sku):
-  product = None
   l_sku = sku.lower().replace(u'-', u'').replace(u' ', u'')
   for a_sku, sku_values in SKUS.items():
-    if l_sku == a_sku.lower().replace(u'-', u'') or l_sku in sku_values[u'aliases']:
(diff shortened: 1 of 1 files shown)


taers232c:
Jay,

Any thoughts on how you'd like to proceed.

Ross

taers232c:
By applying this PR, Gam is internally consistent. My concern is still that users have taken the SKUs from Gam and have used them in external data/apps. By switching to the SKU display name that's going to break. By leaving the existing data alone and added an additional column for print licenses/users, nothing breaks and the user gets the friendlier SKU display name as a bonus.

jay0lee:
The fact is the API is currently in an inconsistent state. For example, Google apps business maps to g suite basic while g suite business maps to google apps unlimited.

I'm not concerned about possibly breaking we admins downstream scripts, they can update them. Right now I'm aiming to have gam aligned with what the admin console diaplays. No further work is necessary here.

PR #437, opened 2017-02-19
License optimization

taers232c:
doPrintLicenses - 9444/9492
When return_list is true, return: `{"primaryEmail1": [list of SKUs], "primaryEmail2": [list of SKUs']}`

getUsersToModify - 9637
The list of users is just the keys from the dictionary

doPrintUsers - 8780/8784
Eliminate scanning through licenses for the userId, just get the SKU list from the dictionary
Convert users SKUs from above to displayable form

doGetUserInfo - 7514, doPrintLicenses - 9454, 9457
Lists of SKUs/products can't be split on space, see getUsersToModify - 9637,

### DIFF
--- a/src/gam.py
+++ b/src/gam.py
@@ -7511,7 +7511,7 @@ def doGetUserInfo(user_email=None):
       getLicenses = False
       i += 1
     elif myarg in [u'sku', u'skus']:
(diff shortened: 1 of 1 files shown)


taers232c:
Will repost after submitting a different PR

PR #446, opened 2017-03-16
Improve API caching options, cache errors

taers232c:
Caching
* If neither nocache.txt or allcache.txt are present, only API discovery calls are cached
* If nocache.txt is present (allcache.txt is ignored), no API calls are cached
* If only allcache.txt is present, all API calls are cached

Clean up error messages
* The temporary error/backoff message is now written on one line, in gam csv operations having it split over the sleep gave muddled output when another process wrote at the same time
* e.message is deprecated, replaced with str(e)
* Move a couple of error messages from stdout to stderr

Code cleanup
* Make function getService(api, http) that gets the API info and gets the service
* Eliminates duplicated code in buildGAPIObject and buildGAPIServiceObject
* Handles invalid cached discovery document by turning caching off

### DIFF
--- a/src/gam.py
+++ b/src/gam.py
@@ -380,8 +380,9 @@ def SetGlobalVariables():
   _getOldSignalFile(GC_DEBUG_LEVEL, u'debug.gam', filePresentValue=4, fileAbsentValue=0)
   _getOldSignalFile(GC_NO_VERIFY_SSL, u'noverifyssl.txt')
   _getOldSignalFile(GC_NO_BROWSER, u'nobrowser.txt')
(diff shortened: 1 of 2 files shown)


jay0lee:
I'd like to have ALL caching off by default. I don't see the need to confuse admins with discovery caching vs regular caching so allcache.txt can turn both on.

taers232c:
Jay,

I'm not sure I understand your comment.

Currently:
	nocache.txt present: no caching
	default: all caching

My proposal:
	nocache.txt present: no caching
	allcache.txt present: all caching
        default: discovery caching

There are real savings by caching the discovery documents, that's my I've made that the default.

Ross

--
ross.scroggs@gmail.com

PR #458, opened 2017-03-25
Fix add smime, isDefault can not be set on insert

taers232c:


taers232c:
From: https://developers.google.com/gmail/api/guides/smime_certs
Upload an S/MIME key

The S/MIME key should be present in the pkcs12 field in that format; no other fields should be set in the request.

PR #484, opened 2017-04-18
Establish GC_Values[GC_DOMAIN] for calendar commands

taers232c:
This now works as `GC_Values[GC_DOMAIN]` is set:
`gam calendar resourceemailaddressXXXX@resource.calendar.google.com update read domain`
This (and similar commands) now properly add the domain:
`gam calendar testuser showacl`

### DIFF
--- a/src/gam.py
+++ b/src/gam.py
@@ -819,6 +819,8 @@ def buildActivityGAPIObject(user):
   return (userEmail, buildGAPIServiceObject(u'appsactivity', userEmail))
 
 def buildCalendarGAPIObject(calname):
+  if not GC_Values[GC_DOMAIN]:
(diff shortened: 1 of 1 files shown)


src/gam.py
```diff
@@ -7742,6 +7745,13 @@ def _getAdminUserFromOAuth():
     credentials = storage.get()
   return credentials.id_token.get(u'email', u'Unknown')
 
+def _getDomainFromOAuth():
```
jay0lee: Let's just take _getAdminUserFromOAuth() and generalize it so something like _getValueFromOAuth. You would specify 'email' or 'hd' (or anything else in credentials) and get that value returned.

PR #517, opened 2017-06-30
Updated gam create group to avoid a Google API issue that generated a…

taers232c:
…n error when the group description contains a <, > or =.

### DIFF
--- a/src/gam.py
+++ b/src/gam.py
@@ -7000,11 +7000,13 @@ def doCreateGroup():
       i += 2
     elif myarg == u'description':
       description = sys.argv[i+1].replace(u'\\n', u'\n')
-      if description.find(u'\n') != -1:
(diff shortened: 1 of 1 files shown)


src/gam.py
```diff
@@ -7000,11 +7000,13 @@ def doCreateGroup():
       i += 2
     elif myarg == u'description':
       description = sys.argv[i+1].replace(u'\\n', u'\n')
-      if description.find(u'\n') != -1:
-        gs_body[u'description'] = description
-        if not gs:
-          gs = buildGAPIObject(u'groupssettings')
-          gs_object = gs._rootDesc
+      for c in u'\n<>=':
```
jay0lee: looks fine but can you add a brief comment just above as to why this is necessary so it will jog our memory in the future?

PR #523, opened 2017-07-01
Fix course state validation

taers232c:


jay0lee:
See patch I just committed. Where the discovery files offer "enum" fields for values I'd prefer we use those since they are dynamic and if Google adds a new state GAM will "just work" with it.

PR #539, opened 2017-07-27
Standardize callGAPIpages calls; cleanup new Vault code

taers232c:
All calls to callGAPIpages include the items parameter; none of them omit it and take the default value u'items'; thus items can be a positional parameter and the items= in each of the calls can be omitted. Previously, some calls had item= and others didn't.

Cleaned up Vault argument processing.

Clean up some pylint complaints.

### DIFF
--- a/src/gam.py
+++ b/src/gam.py
@@ -7250,11 +7250,11 @@ def doGetVaultHoldInfo():
       print u'ERROR: %s is not a valid argument for "gam info hold"' % myarg
       sys.exit(3)
   results = callGAPI(v.matters().holds(), u'get', matterId=matterId, holdId=holdId)
-  buildGAPIObject(u'directory')
+  cd = buildGAPIObject(u'directory')
   if u'accounts' in results:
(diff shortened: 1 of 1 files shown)


src/gam.py
```diff
@@ -7432,7 +7434,7 @@ def doGetVaultMatterInfo():
   matterId = convertMatterNameToID(v, sys.argv[3])
   result = callGAPI(v.matters(), u'get', matterId=matterId, view=u'FULL')
   if u'matterPermissions' in result:
-    cd = buildGAPIObject(u'directory')
+    buildGAPIObject(u'directory')
```
jay0lee: I think this was a mistake on my part, should have built cd here and then passed it to:

user_email = convertUserUIDtoEmailAddress(

further down so that we don't waste cycles on building directory API object every time through the loop

taers232c:
Fixed in doGetVaultHoldInfo and doGetVaultMatterInfo

PR #550, opened 2017-08-17
Add sharedparentname to add/update drivefile

taers232c:
This allows the user to select a parent folder other that one she owns, a shared folder typically

Documentation updates


taers232c:
No progress here, back to vault updates

PR #558, opened 2017-08-26
Display correct owner on Drive file lists

flowerysong:
Because the default search query is "'me' in owners", the previous code
assumes that the owner is a static value. This assumption is broken when
"anyowner" or "fullquery" are used. We should instead fetch the actual
ownership information and use that.

### DIFF
--- a/src/gam.py
+++ b/src/gam.py
@@ -3966,17 +3966,15 @@ def printDriveFileList(users):
       sys.exit(2)
   if fieldsList or labelsList:
     fields = u'nextPageToken,items('
-    if fieldsList:
-      fields += u','.join(set(fieldsList))
-      if labelsList:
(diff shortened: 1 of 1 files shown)


taers232c:
flowerysong,

Yes, in the old code the only files you could see were owned by the specified user, so the static Owner value made sense.
Now that you can select files not owned by the specified user, the column should probably be named User, but that change would break a lot of scripts.
The values in the rest of the columns are relative to the user named in the Owner(User) column, not the actual owner of the file (unless they are the same).
Putting the actual owner in the Owner(User) column would make some of the other columns invalid if the specified user and owner are not the same.
The owners.0.emailAddress column gives you the actual owner.

Ross
--
ross.scroggs@gmail.com

PR #568, opened 2017-09-11
Add INTER_BATCH_WAIT to allow batch speed limiting

taers232c:


jay0lee:
With adding / removing members from groups, have you done any performance testing to see if the batch API calls actually improves performance (and doesn't hit API limits) over the old code?

PR #570, opened 2017-09-14
Add inter_batch_wait

taers232c:
Can we go ahead and enter this now to give sync users a solution that works with gam batch and does paced adds/remove. Nothing prevents the users from doing: gam csv 1000users.csv gam update group oldgam add member ~email

### DIFF
--- a/src/gam.py
+++ b/src/gam.py
@@ -386,6 +386,7 @@ def SetGlobalVariables():
   _getOldEnvVar(GC_ACTIVITY_MAX_RESULTS, u'GAM_ACTIVITY_MAX_RESULTS')
   _getOldEnvVar(GC_AUTO_BATCH_MIN, u'GAM_AUTOBATCH')
   _getOldEnvVar(GC_BATCH_SIZE, u'GAM_BATCH_SIZE')
+  _getOldEnvVar(GC_INTER_BATCH_WAIT, u'GAM_INTER_BATCH_WAIT')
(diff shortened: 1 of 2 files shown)


jay0lee:
I don't see the point of adding further complexity when batch API calls is slower than parallel API calls. Can we look at re-implementing with parallel API calls?

taers232c:
Parallel add/removes have not been eliminated, you can still do that if you want.
What batch does is speeds up: gam update group group@domain.com <mailto:group@domain.com> add/remove/sync member <ou /Path/To/OU, file FileName, ...>

--
ross.scroggs@gmail.com

jay0lee:
but it's not actually faster than parallel. I'm looking at re-implementing add/remove/sync to use parallel.

taers232c:
You have to update your school groups every night to allow for new students/departing students.
On the first day of school, batch may be somewhat slow that parallel, but on every other night you pay a big price.
Batch:
gam update group seniors@domain.com <mailto:seniors@domain.com> sync members ou /Students/ClassOf2018

or 

Parallel:
gam update group seniors@domain.com <mailto:seniors@domain.com> clear member
echo "primaryEmail" > users.csv
gam ou /Students/ClassOf2018 print >> users.csv
gam csv users.csv gam update group seniors@domain.com <mailto:seniors@domain.com> add member ~primaryEmail
--
ross.scroggs@gmail.com

jay0lee:
See my commit just now to switch to using parallel instead of batch for add/remove/sync. It still needs some cleanup but I'm going to stick with parallel unless you can convince me that batch is either faster or simpler.

PR #571, opened 2017-09-14
Minor locations/sshPublicKeys cleanup

taers232c:
Make options names internally consistent and consistent with other commands.

Drop publicsshkeys, too may choices

Change title to SSH Public Keys to match Google's title

### DIFF
--- a/src/GamCommands.txt
+++ b/src/GamCommands.txt
@@ -380,13 +380,16 @@ Named items
 	is2svenforced|
 	is2svenrolled|
 	lastlogintime|
+	locations|location|
 	noneditablealiases|aliases|nicknames|
 	notes|note|
 	organizations|organization|
 	orgunitpath|org|ou|
 	phones|phone|
+	posixaccounts|posix|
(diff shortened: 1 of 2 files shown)


src/GamCommands.txt
```diff
@@ -380,13 +380,16 @@ Named items
...
+	locations|location|
 	noneditablealiases|aliases|nicknames|
 	notes|note|
 	organizations|organization|
 	orgunitpath|org|ou|
 	phones|phone|
+	posixaccounts|posix|
 	primaryemail|username|
 	relations|relation|
+	ssh|sshkeys|sshpublickeys|
```
jay0lee: you could probably simplify this doc file a lot by not listing every possible argument that performs the same actions and instead listing the "best" choice.

taers232c:
I'll edit it if you let me know what you believe the "best" choices are.
--
ross.scroggs@gmail.com

taers232c:
Did some simplifications.

PR #576, opened 2017-09-16
Update doUpdateCros

taers232c:
Only use moveDevicesToOu if the only field being updated is orgUnitPath; otherwise orgUnitPath will be updated in the loop with whatever other fields are being updated

Update GamCommand for PR #574


jay0lee:
We should always be using moveDevicesToOU, even if it's just one device, update should not be used for our moves.

taers232c:
If you are updating multiple fields, then I just use the update loop. Why do one loop to update the other fields and then use moveDevicesToOu to update the OUs.

Is updating the OU in update broken?

jay0lee:
movdDevicesToOU is ultimately faster and more efficient for the devices themselves to get updated. It should be used for all device OU moves now.

PR #586, opened 2017-09-17
Handle case where several user attribute items are set primary

taers232c:
If an item has a primary field and it's set True, set primary to False in all previous items where primary is set True

Otherwise, the API returns the error: "Invalid Input: Bad request for -" which doesn't explain that multiple items had primary True


jay0lee:
If admin is specifying two orgs, phones, etc as primary, shouldn't we throw an error here? That allows admin to correct which one is primary without us making an arbitrary choice.

PR #589, opened 2017-09-18
Clean up customer subscription code

taers232c:


src/gam.py
```diff
@@ -8330,26 +8330,23 @@ def doWhatIs():
...
-  results = callGAPI(res.subscriptions(), u'list', customerId=customerId)
-  for subscription in results[u'subscriptions']:
-    if sku == subscription[u'skuId']:
-      subscriptionId = subscription[u'subscriptionId']
-      break
-  if not subscriptionId:
-    print u'ERROR: could not find subscription for customer %s and SKU %s' % (customerId, sku)
-    sys.exit(3)
-  return subscriptionId
+def getCustomerSubscription(res):
```
jay0lee: What is the advantage of these changes?

taers232c:
doDeleteResoldSubscription and doUpdateResoldSubscription take the same three arguments so I make s subroutine that gets them as a unit.

taers232c:
I'll put it back your way

src/gam.py
```diff
@@ -8330,26 +8330,23 @@ def doWhatIs():
...
+  customerId = sys.argv[3]
+  skuId = sys.argv[4]
```
jay0lee: I don't like these utility functions depending on sys.argv order, it means they can't be used later if argument order is different. I think convertSKU2ProductId() is pretty much fine as is though the argument cleanup in _getResoldCustomerAttr() is good.

PR #590, opened 2017-09-18
Simpler clean up subscriptions

taers232c:


taers232c:
Is this more like what you had in mind?

PR #592, opened 2017-09-19
Update update group

taers232c:
Allow plurals of roles, otherwise bad things happen; for example, in the following members isn't taken as the role, it's taken as the list of members (file file.txt is ignored) and the group is trashed.
gam update group group@domain.com sync members file file.txt

In sync, handle googlemail.com as a synonym for gmail.com and dots in googlemail.com/gmail.com addresses. For both the incoming list of members and the current members, the addresses are normalized and then the adds/remove are determined. The actual address is used in the add/remove.

### DIFF
--- a/src/gam.py
+++ b/src/gam.py
@@ -7909,7 +7909,7 @@ GROUP_ROLES_MAP = {
 def doUpdateGroup():
 
 # Convert foo@googlemail.com to foo@gmail.com; eliminate periods in name for foo.bar@gmail.com
-  def _cleanAddress(emailAddress, mapCleanToOriginal):
+  def _cleanConsumerAddress(emailAddress, mapCleanToOriginal):
     atLoc = emailAddress.find(u'@')
     if atLoc > 0:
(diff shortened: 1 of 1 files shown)


src/gam.py
```diff
@@ -7900,15 +7900,31 @@ def checkGroupExists(cd, group, i=0, count=0):
...
+GROUP_ROLES_MAP = {
+  u'owner': ROLE_OWNER, u'owners': ROLE_OWNER,
+  u'manager': ROLE_MANAGER, u'managers': ROLE_MANAGER,
+  u'member': ROLE_MEMBER, u'members': ROLE_MEMBER,
+  }
 
 def doUpdateGroup():
 
+# Convert foo@googlemail.com to foo@gmail.com; eliminate periods in name for foo.bar@gmail.com
+  def _cleanAddress(emailAddress, mapCleanToOriginal):
```
jay0lee: Maybe _cleanConsumerAddresss() ?

PR #594, opened 2017-09-19
Misc updates

taers232c:
doUpdateCustomer - skip API call if user didn't update anything

getPhoto - handle uknkown user and user with no photo

doLabel, updateLabels - cleanup, make consistent

doGetGroupInfo - set settings in all cases to avoid having to use try/except

doPrintGroupMembers - message_attributes not used

### DIFF
--- a/src/gam.py
+++ b/src/gam.py
@@ -1706,7 +1706,10 @@ def doUpdateCustomer():
       sys.exit(2)
   if body:
     callGAPI(cd.customers(), u'update', customerKey=GC_Values[GC_CUSTOMER_ID], body=body)
-  print u'Updated customer'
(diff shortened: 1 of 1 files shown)


src/gam.py
```diff
@@ -1704,7 +1704,8 @@ def doUpdateCustomer():
     else:
       print u'ERROR: %s is not a valid argument for "gam update customer"' % myarg
       sys.exit(2)
-  callGAPI(cd.customers(), u'update', customerKey=GC_Values[GC_CUSTOMER_ID], body=body)
+  if body:
```
jay0lee: I'd say throw an error that no update was specified and exit.

PR #595, opened 2017-09-19
When displaying license info to user, show Id and displayName if available

taers232c:


taers232c:
Are you going to include this one in 4.31?

PR #644, opened 2017-12-16
Enhance device file handling

taers232c:
In info cros, allow downloadfile selection by name or time.
Download folder defaults to GAMDRIVEDIR, can be overridden with targetfolder

_checkTPMVulnerability doesn't need to return cros as it's updated in place

Allow device files to be handled like recent users and time ranges in print cros/crosactivity

### DIFF
--- a/src/gam.py
+++ b/src/gam.py
@@ -9193,7 +9193,8 @@ def doGetCrosInfo():
       devices.append(a_device[u'deviceId'])
   else:
     devices = [deviceId,]
-  downloadfile = None
+  downloadFileFieldName = downloadFileFieldValue = None
+  targetFolder = GC_Values[GC_DRIVE_DIR]
   projection = None
   fieldsList = []
   noLists = False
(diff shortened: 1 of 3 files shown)


src/gam.py
```diff
@@ -7970,21 +7970,21 @@ def _getBuildingAttributes(args, body={}):
...
-  body = _getBuildingAttributes(sys.argv[3:], body)
+  _getBuildingAttributes(sys.argv[3:], body)
   print u'Creating building %s...' % body[u'buildingId']
   callGAPI(cd.resources().buildings(), u'insert',
            customer=GC_Values[GC_CUSTOMER_ID], body=body)
 
 def doUpdateBuilding():
   cd = buildGAPIObject(u'directory')
   buildingId = sys.argv[3]
-  body = _getBuildingAttributes(sys.argv[4:])
```
jay0lee: I understand the behavior is the same here but the assignment ( body = ) makes it obvious that the function is actually modifying body. Let's stick with that form here.

src/gam.py
```diff
@@ -7974,17 +7974,15 @@ def _getBuildingAttributes(args, body={}):
 
 def doCreateBuilding():
   cd = buildGAPIObject(u'directory')
-  body = {u'floorNames': [u'1'],
-          u'buildingId': unicode(uuid.uuid4())}
-  body = _getBuildingAttributes(sys.argv[3:], body)
+  body = _getBuildingAttributes(sys.argv[3:], {u'floorNames': [u'1',], u'buildingId': unicode(uuid.uuid4())})
```
jay0lee: Let's keep this multi-line for easier readability. Let's also default body to {} in the function def.

PR #648, opened 2017-12-18
Make create/update building consistent

taers232c:
Make create building match create resource. They both require an immutable ID and a mutable name.
```
  body = {u'buildingId': sys.argv[3],
          u'buildingName': sys.argv[4],
          u'floorNames': [u'1',]}
  body = _getBuildingAttributes(sys.argv[5:], body)
 ```
```
  body = {u'resourceId': sys.argv[3],
          u'resourceName': sys.argv[4]}
  body = _getResourceCalendarAttributes(sys.argv[5:], body)
```

### DIFF
--- a/src/GamCommands.txt
+++ b/src/GamCommands.txt
@@ -878,7 +878,7 @@ gam update notification|notifications [(id all)|(id <NotificationID>)*] unread|r
 gam delete notification|notifications [(id all)|(id <NotificationID>)*]
 gam info notification|notifications [unreadonly]
 
-gam create building <BuildIngID> <BuildingAttributes>*
+gam create building <BuildIngID> <Name> <BuildingAttributes>*
 gam update building <BuildIngID> <BuildingAttributes>*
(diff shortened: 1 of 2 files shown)


jay0lee:
Ross, thanks for pointing the name/id inconsistency out. I just pushed an update that allows specifying name or id for update/get/delete building commands. A bit of my logic here:

- spoke with team launching this functionality and apparently buildingId will generally not be shown in Calendar UIs, it's meant as a unique identifier.
- Since it's not exposed to UI, we should not force admin to generate a unique ID when creating the resource. We'll default to using a randomized uuid for ID but also allow admin to set ID at create time if they choose to do so.
- We'll assume name is what admin is specifying on update/info/delete commands unless it's prefixed with "id:". This means we'll need to pull a full list of buildings to find the correct building and also handle case insensitivity issues (e.g. buildings named "Empire State Building" and "empire state building" can exist at same time.

PR #653, opened 2017-12-19
It's not a user error in the API gives us a bogus buildingId

taers232c:


jay0lee:
Have you gotten the API to return bogus id some way? User error or no we should probably flag it as some sort of failure rather than returning I'd.

PR #654, opened 2017-12-20
Final building/feature/resource calendar cleanup

taers232c:
6905 - buildingId in user does not seem to refer to buildings
8026 - simplify
11284 - 11300 - Allow user to select new resoure calendar fields
11324, 11372 - make callGAPIpages calls similar to all others
11432 - collapse features to a list of names so it can be fed back to update resource
11438 - Organize output

### DIFF
--- a/src/GamCommands.txt
+++ b/src/GamCommands.txt
@@ -118,7 +118,7 @@ Named items
 <ASPID> ::= <String>
 <BuildingID> ::= <String>|id:<String>
 <CalendarACLRole> ::= editor|freebusy|freebusyreader|owner|reader|writer
-<CalendarColorIndex> ::== <Number in range 1-24>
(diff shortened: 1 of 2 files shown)


jay0lee:
What issues have you seen with buildingId for users?

taers232c:
You can put any value you want, it is not checked as being an actual building id. If you put an invalid buildingId in a resource, the API throws an error.
Is it supposed to refer to an actual building?
--
ross.scroggs@gmail.com

jay0lee:
Yes, it should though I think for user it's not strictly enforced.

The whole purpose of buildings/floors/sections is to allow Calendar to determine which conference rooms are close to a user based on the user's location and the conference room's.

Even if API is not strict about accepting a valid buildingId I believe GAM should be.

PR #662, opened 2017-12-21
Show deviceFiles.downloadUrl in print cros

taers232c:


jay0lee:
I'm not sure what point of showing the URL is here. You can't use curl or wget to download the file since it needs to be authorized.

PR #721, opened 2018-04-16
Add user agent header wrapper to requests handled by AuthorizedHttp

ejochman:
Make the existing solution a bit more flexible by passing a method to be wrapped.


taers232c:
Very nice, thank you.

Ross

PR #744, opened 2018-07-04
Updates

taers232c:
973-1008 - Internally cache discovery documents; speeds up loops over users

1857 - Use patch instead of update so customer address doesn't get wiped out

9917 - In writeCSVFiles, the number of columns is the number of titles. The number of columns in the first row may be smaller than that, think a file with one ACL where other ffiles have many

7171,7173, 10550, 10886, 11655, 12014, 12017 - Standardize error handling

### DIFF
--- a/src/gam.py
+++ b/src/gam.py
@@ -970,10 +970,17 @@ def getValidOauth2TxtCredentials():
 
 def getService(api, http):
   api, version, api_version = getAPIVersion(api)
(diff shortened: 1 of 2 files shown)


jay0lee:
The patch is removing g suite Enterprise for education product SKU, assume that's an accident?

taers232c:
Yes, fixed
--
ross.scroggs@gmail.com

Issue #597, opened 2017-09-20
Add Failed: 403: quotaExceeded - Request rate higher than configured.


rg954:
The issue tracker is for reporting product deficiencies. "How do I?" questions should be posted to the discussion forum at https://groups.google.com/group/google-apps-manager. When in doubt, start at the discussion forum and return here only when instructed to do so.

Please confirm the following:
* I have upgraded to the latest GAM release from https://git.io/gamreleases and I still have this issue.
* I am typing the command as described in the GAM Wiki at https://github.com/jay0lee/gam/wiki

Full steps to reproduce the issue:
1.  Enter Command 

> gam update group (group email) add member org (OU Path)

2.
3.

Expected outcome (what are you trying to do?): Update my group with the members of the current OU. I run this every so often to make sure it is up to date with the most recently created users. It typically goes through the entire user list of users in the OU, and will return a "Error: 409: Member already exists" if they are already in the group, but then adds those that aren't. 

Actual outcome (what errors or bad behavior do you see instead?): Instead of adding the new members, I received an "Add Failed: 403: quotaExceeded - Request rate higher than configured." I attributed it to the new batch feature in 4.30. I downgraded to 4.23 (the last version I used to update the groups?, and the command ran without issue.

lucasndi:
Same issue here, 
Acording [API documentation](https://developers.google.com/admin-sdk/directory/v1/limits)
 [this ](https://developers.google.com/admin-sdk/directory/v1/limits#backoff)can help : 

My workarround is to run it many times

ntsafnat:
Same issue here:

RuntimeError: maximum recursion depth exceeded
[36188] Failed to execute script gam

using GAM 4.12 with the same command works fine.

jay0lee:
Closing as this should be fixed in current 4.40 release.

PR #751, opened 2018-07-04
Use batch for group add/remove

taers232c:
This allows update group add/remove/sync/clear to be done from gam batch or gam csv


jay0lee:
For reference, it was #597 (note time of year)

PR #753, opened 2018-07-05
Make collaborative consistent

taers232c:


jay0lee:
managers mapping to managers and owners setting is intentional as group roles are intended to be additive. ALL_MANAGERS excluding the right from owners is not intuitive and owners can just give themselves the right anyway. Let's allow the setting with managersonly but it should generally not be used.

PR #757, opened 2018-07-11
Add owneremail option to print courses

taers232c:
This will get owner email from owner ID


taers232c:
I've got some small fixes in the wings, do you want to process this?

PR #770, opened 2018-08-03
Add textcolor and backgroundcolor to add label/update labelsettings

taers232c:


jay0lee:
Ideally we'd map the HTML color codes to simpler color names like red, green, etc to make life easier for the admin but that can come at a later time.

PR #788, opened 2018-08-29
If language missing, default to en. Fixes #787

### LINKED ISSUE #787: gam info domain - KeyError: u'language'
Full steps to reproduce the issue:
1. gam info domain

Actual outcome (what errors or bad behavior do you see instead?):
<code>
  File "/Users/jan/R-D/gam/src/gam.py", line 13170, in <module>
    sys.exit(ProcessGAMCommand(sys.argv))
  File "/Users/jan/R-D/gam/src/gam.py", line 12575, in ProcessGAMCommand
    doGetDomainInfo()
  File "/Users/jan/R-D/gam/src/gam.py", line 1722, in doGetDomainInfo
    doGetCustomerInfo()
  File "/Users/jan/R-D/gam/src/gam.py", line 1752, in doGetCustomerInfo
    print u'Default Language: %s' % customer_info[u'language']
KeyError: u'language'
</code>

I checked the response for this client from the [API](https://developers.google.com/admin-sdk/directory/v1/reference/customers/get). It is actually not returning any language-key, so I suppose it should default to 'en' as stated in the [overview](https://developers.google.com/admin-sdk/directory/v1/reference/customers).

jalmeroth:


jay0lee:
lets make it clear that it was unset, change en to something like:

Unset (defaults to en)

jalmeroth:
Like this?

PR #808, opened 2018-10-15
Add operatingSystemType to posix config

### LINKED ISSUE #807: Directory API: operatingSystemType parameter can not be specified for POSIX accounts
The `posixAccounts` of a user has a field, `operatingSystemType` which may be used to limit user accounts to a specific type.

Here is an example of an organically generated `posixAccounts` entry

```
Posix Accounts:
 username:  username_domain_tld
 uid: 1696033707
 primary: True
 operatingSystemType: linux
 gid: 1696033707
 systemId: 
 homeDirectory: /home/username_domain_tld
```

Suppose that I update this entry such that the user's name is `username` rather than `username_domain_tld` by running `gam update user username@domain.tld posixaccounts uid 1696033707 gid 1696033707 username username endposix`

The changes are reflected in the updated `gam info` output -

```
Posix Accounts:
 username: username
 uid: 1696033707
 primary: True
 operatingSystemType: unspecified
 gid: 1696033707
 homeDirectory: /home/username
```

Notice that `operatingSystemType` is `unspecified`.

Investigation of the GAM source code would suggest that GAM has no concept of the operatingSystemType parameter.

This is an important feature and should be added.

RomanHargrave:
Closes #807 

This adds a parameter `os` (or `operatingSystemType`) to `posixaccounts`

### DIFF
--- a/src/gam.py
+++ b/src/gam.py
@@ -6930,7 +6930,7 @@ def getUserAttributes(i, cd, updateCmd):
         elif myopt in [u'username', u'name']:
           posix[u'username'] = sys.argv[i+1]
           i += 2
-        elif myopt in [u'os', u'operatingSystemType']:
+        elif myopt in [u'os', u'operatingsystemtype']:
           posix[u'operatingSystemType'] = sys.argv[i+1]
           i += 2
         elif myopt in [u'endposix']:


taers232c:
Roman,

This line: elif myopt in [u'os', u'operatingSystemType']:
should be: elif myopt in [u'os', u'operatingsystemtype']:

myopt is in lower case

Ross

RomanHargrave:
@taers232c applied

PR #822, opened 2018-11-25
Handle no dns.resolver

taers232c:


jay0lee:
we should probably handle dnspython the same way we currently handle all other libraries and just pull it into our tree here.At some point I'd like to get things pip installable and working more like a regular Python app repository but that's a larger issue.

PR #827, opened 2018-12-18
Simplify gam version

taers232c:


jay0lee:
The goal here was to only see pre-releases if it's a force check. This allows pushing "beta-ish" features without notifying everyone. I haven't used it in awhile but would like to keep it so that scheduled checks don't see pre-release versions but forced checks do.

taers232c:
OK

PR #830, opened 2018-12-28
Allow updating ChromeOS device notes while retaining existing value

taers232c:
This allows you to combine new notes with existing notes:
```
gam update cros <CrOSEntity> updatenotes "New note/#note#"
```


jay0lee:
I think I'd rather just see the old notes appended automatically rather than requiring admin to put #note# in the value. If they want more flexibility over the note separators they can already do:

gam print cros notes query id:<SN> | gam csv - gam update cros ~id notes "new note/~~notes~~"

or such.

PR #835, opened 2019-01-08
Modified several commands to set environment variables with the ID of the item created in the command

taers232c:
This allows use of the ID in subsequent commands via the environment variable


jay0lee:
Have you tested this on various platforms? I believe in Linux at least, child processes cannot set environment variables for their parent process. It may work within the context of Gam batch though. Just curious what you found.

taers232c:
Dang! Seemed like a good idea. Thanks

PR #850, opened 2019-02-18
Remove extraneous files/folders

taers232c:
Remove *.pyi in dns
Remove duplicate oauth2client inside oauth2client


jay0lee:
Let's leave the .pyi files as they are a part of the pydns distribution.

PR #861, opened 2019-03-08
Disable POP and IMAP when deprovisioning user

pawelpbm:


jay0lee:
in general, I'm moving away from these "multi action" commands. Instead admins should work on building a batch of commands that they wish to be executed in a given sequence for user actions.

PR #878, opened 2019-04-08
Toss some more Google+, update V1_DISCOVERY_APIS

taers232c:


src/var.py
```diff
@@ -99,17 +99,18 @@
...
 V1_DISCOVERY_APIS = {
-  u'oauth2',
-  u'reseller',
-  u'storage',
-  u'drive',
   u'admin',
+  u'appsactivity',
+  u'calendar',
+  u'drive',
+  u'drive3',
```
jay0lee: this shouldn't' be necessary as we should have already switched to the true API name / version by now. This is the case with "admin" instead of directory / reports / datatransfer. Are you seeing differently?

PR #879, opened 2019-04-09
New limit on number of cells

taers232c:
https://support.google.com/drive/answer/37603

### DIFF
--- a/src/gam.py
+++ b/src/gam.py
@@ -10638,6 +10638,7 @@ and follow recommend steps to authorize GAM for Drive access.''' % (admin_email)
     cell_count = rows * columns
     data_size = string_file.len
     max_sheet_bytes = int(result[u'maxImportSizes'][MIMETYPE_GA_SPREADSHEET])
+# See https://support.google.com/drive/answer/37603 for limits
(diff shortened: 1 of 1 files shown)


src/gam.py
```diff
@@ -10638,7 +10638,7 @@ def writeCSVfile(csvRows, titles, list_type, todrive):
     cell_count = rows * columns
     data_size = string_file.len
     max_sheet_bytes = int(result[u'maxImportSizes'][MIMETYPE_GA_SPREADSHEET])
-    if cell_count > 2000000 or data_size > max_sheet_bytes:
+    if cell_count > 5000000 or data_size > max_sheet_bytes:
```
ejochman: Can you add a comment here with the link in the description, so as to add context to this limit? It's otherwise very opaque where these [magic numbers](https://en.wikipedia.org/wiki/Magic_number_(programming)) come from.

jay0lee: In addition to the link comment we should probably make this a variable in vars.py that can be overridden. That way next time there's a bump users can set the new value without the need to upgrade (or wait for us to push a new GAM version).

PR #882, opened 2019-04-11
Refactor scope selection menu

ejochman:
-Add flexibility to menu creation and feature customization
-Colorize menu error messages on supported platforms
-Add 'email' as a required scope for increased transparency to the user
-Improve readability of menu creation and operation

### DIFF
--- a/src/gam.py
+++ b/src/gam.py
@@ -115,7 +115,7 @@ def _request_with_user_agent(request_method):
   return wrapped_request_method
 
 google_auth_httplib2.Request.__call__ = _request_with_user_agent(
-    google_auth_httplib2.Request.__call__)	
+    google_auth_httplib2.Request.__call__)
 google_auth_httplib2.AuthorizedHttp.request = _request_with_user_agent(
     google_auth_httplib2.AuthorizedHttp.request)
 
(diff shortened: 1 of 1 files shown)


taers232c:
In this code:
```
  def select(self, restriction=None):
    """Selects/enables the ScopeMenuOption with an optional restriction."""
    if restriction is not None:
      self.restrict_to(restriction)
    self.is_selected = True

  def unselect(self):
    """Unselects/disables the ScopeMenuOption."""
    self.is_selected = False

  @property
  def is_selected(self):
    return self._is_selected

  @is_selected.setter
  def is_selected(self, is_selected):
    if self.is_required and not is_selected:
      raise ValueError('Required scope cannot be unselected')
    if not is_selected:
      # Disable all applied restrictions                                                                                                                                                                                                                                                                                          
      self.unrestrict()
    self._is_selected = is_selected
```
what is the relationship between `self.is_selected` and `self._is_selected`?
pylint complains:
```
W:12755, 4: Attribute '_is_selected' defined outside __init__ (attribute-defined-outside-init)
```

ejochman:
The linter is right. I should have initialized those private members in
__init__ as a best practice. I'll send a followup to do that.

As far as the difference between `_is_selected` and `is_selected`:
`_is_selected` is the private member and its job is strictly to hold the
final boolean value of whether the item is currently selected or not; while
is_selected is the public property that exposes the value. Making it a
property allows us to add some logic and enforcement to how the raw value
gets set, whereas just assigning a value to the member does not. We could
use actual methods like setIsSelected() and getIsSelected(), but the
property functionality is more multipurpose in this application.

I didn't read it too closely, but it looks like this page has a decent
explanation: https://www.python-course.eu/python3_properties.php

taers232c:
I deleted my comment after reading up on `@property` and `@x.setter`. The examples I saw looked like yours, they used the setter `self.is_selected = True` in `__init__`. Does pylint want you to say `self._is_selected = True` in `__init--`?

ejochman:
I think it's just complaining that I didn't declare _is_selected on the
class during __init__, but rather it was declared and initialized in the
is_selected setter which was called from __init__. It's a readability
thing, more than anything. It's a lot easier for readers to parse through
your __init__ to understand all the private members that will be available
within the class, rather than having to read through and understand the
entire internal class code to get that same context

PR #883, opened 2019-04-12
Fix bug in setting scopes

taers232c:
If you set an item to read only, you can't get back to not read only
```
21 *->blank
21 blank->*
21r *-> R
21 R->blank
21 blank -> R Should be *
```

### DIFF
--- a/src/gam.py
+++ b/src/gam.py
@@ -13126,6 +13126,7 @@ Append an 'r' to grant read-only access or an 'a' to grant action-only access.
       if selected is None:
         # Toggle the option on/off
         option.is_selected = not option.is_selected
+        option._restriction = None
       else:
         option.is_selected = selected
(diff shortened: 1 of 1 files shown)


src/gam.py
```diff
@@ -13126,6 +13126,7 @@ def _update_option(self, option, selected=None, restriction=None):
       if selected is None:
         # Toggle the option on/off
         option.is_selected = not option.is_selected
+        option._restriction = None
```
ejochman: Please use the accessor here, instead of setting a private member: option.unrestrict()

ejochman: Also, it looks like this restriction removal should be happening automatically in the is_selected property setter, when is_selected is set to False, so I'll have to look into why it doesn't seem to be working correctly.

PR #897, opened 2019-04-22
pylint cleanup, Python 3 cleanup/fixes

taers232c:


src/gam.py
```diff
@@ -32,7 +32,7 @@
 import csv
 import datetime
 import hashlib
-import http.client
+import http.client as http_client
```
jay0lee: thanks for catching. This is fine for now though we should probably fix the issue by not using http as a var name at some point.

src/gam.py
```diff
@@ -3859,11 +3816,10 @@ def getPhoto(users):
       photo_data = photo['photoData']
       if showPhotoData:
         print(photo_data)
-      photo_data = base64.urlsafe_b64decode(photo_data)
     except KeyError:
       print(' no photo for %s' % user)
       continue
-    writeFile(filename, photo_data, mode='wb', continueOnError=True)
+    writeFile(filename, base64.urlsafe_b64decode(photo_data), mode='wb', continueOnError=True)
```
jay0lee: let's keep this as two lines to make it easier to read and under 80 char length.

PR #900, opened 2019-04-24
Code cleanup

taers232c:


jay0lee:
Travis builds seem to fail with an error about default charset:

https://travis-ci.org/jay0lee/GAM/jobs/523802116#L548

Can you take a look?

PR #912, opened 2019-05-02
Gate usage on minimum supported Python version

ejochman:
3.7 is recommended, but 3.5 should be currently supported

### DIFF
--- a/src/gam.py
+++ b/src/gam.py
@@ -14230,6 +14230,6 @@ def ProcessGAMCommand(args):
 if __name__ == "__main__":
   if sys.platform.startswith('win'):
     freeze_support()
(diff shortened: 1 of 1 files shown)


jay0lee:
I ran into a few issues with 3.5. None critical but could see issues down the road:

- json.loads will take bytes in 3.6+, 3.5 requires str https://docs.python.org/3/library/json.html
- Python 3.7 (and OpenSSL 1.1) are needed for the TLS version requirements. It was alarming to see httplib2 using TLSv1 in some cases with older Python and OpenSSL.

Can we look at getting Travis to test 3.5 and 3.6 also (just test, not build them from source)?

jay0lee:
Merging for now, I'll see about adding Python 3.5 and 3.6 Travis testing at some point.

PR #914, opened 2019-05-04
Python 3.5/6 don't have the version attributes

taers232c:


jay0lee:
Actually, if admin says they want to require TLS vN and we don't support forcing that, we should exit with a clear error explaining it requires Python 3.7+ and OpenSSL 1.1+. We should not silently ignore the max/min variables.

PR #918, opened 2019-05-08
Update encoding so Python REs with \ can be processed

taers232c:
Fix bug in printShowDelegates


jay0lee:
latest is failing to compile, is indentation off maybe?

    Syntax error in gam.py
      File "/home/travis/build/jay0lee/GAM/src/gam.py", line 10849
         if op == '<':

PR #919, opened 2019-05-08
Use explicit relative imports for Python 3

ejochman:
In Python 3, implicit relative imports within packages are no longer
available - only absolute imports and explicit relative imports are
supported. In addition, star imports (e.g. from x import *) are only
permitted in module level code.

See: https://www.python.org/dev/peps/pep-0404/#imports

### DIFF
--- a/src/gam.py
+++ b/src/gam.py
@@ -66,8 +66,8 @@ import oauth2client.file
 import oauth2client.tools
 from oauth2client.contrib.dictionary_storage import DictionaryStorage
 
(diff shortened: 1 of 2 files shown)


jay0lee:
Just FYI the travis build failures are due to the .travis script not accounting for PR builds that don't get secure environment variables. I'm working to skip tests that need those variables. That way PRs can at least confirm GAM is still able to run and build to a PyInstaller executable.

PR #938, opened 2019-05-27
Handle missing credentials, e.g., two gam oauth deletes in a row

taers232c:


jay0lee:
showdetails seems unnecessary. Let's just parse out anything in there other than kind, etag, etc.

PR #947, opened 2019-06-07
Handle unknown server in gam version extended location

taers232c:


jay0lee:
was literally just starting on this... :-)

PR #954, opened 2019-06-18
Update var.py

taers232c:
622/3 - I assume that this is a typo on the web page as it is the only entry without a space in front of the (
634 - Duplicate


jay0lee:
couple things here:
- changes need to be made in tools/parse-aue.py so we can continue to parse the Google help article easily as Google adds new models and makes other changes to the page.
- I'm about to submit a change to use a dict in parse-aue.py so we don't get dups (and the code is cleaner)
- I'm going to move the AUE dates list out of var.py and into a standalone file that GAM will download from GitHub on demand. That will allow updates to AUE without the need to upgrade GAM while also giving us a chance to confirm Google didn't modify the help article HTML in some way that trashes parse-aue.py output. It'll also make updates easier since we can just run `parse-aue.py > cros-aue-dates.txt`, `git diff` to see changes/do sanity checks and then push it.
- Google's [recovery.json](https://dl.google.com/dl/edgedl/chromeos/recovery/recovery.json) is useful when comparing model names between the AUE page and what the API / admin console return. Model names from API should generally match model names in the recovery.json file though due to timing issues a model may be in one and not the other. Based on recovery.json, the cb5-311 will show as "Acer Chromebook 13 (CB5-311)" and there is no c810. Since this model is reaching AUE in September c810 is likely just a marketing name by Acer that is really just the CB5-311.

PR #961, opened 2019-06-25
Clean up sendOrDropEmail

taers232c:


src/gam.py
```diff
@@ -5346,8 +5346,12 @@ def transferDriveFiles(users):
 
 def sendOrDropEmail(users, method='send'):
   body = subject = ''
-  recipient = labels = sender = None
+  recipient = sender = None
+  labels = []
```
jay0lee: We should keep this as None so that if labels are not set we don't set the labels argument on the API call at all.

src/gam.py
```diff
@@ -11051,6 +11050,16 @@ def sortCSVTitles(firstTitle, titles):
   for title in restoreTitles[::-1]:
     titles.insert(0, title)
 
+def QuotedArgumentList(items):
```
jay0lee: - we need to escape the text also in case the user ran something like:

gam print users query "abc=\"def\""

- instead of qstr and needing to remove the last space, make a qarray and return the join.

taers232c: You got lucky with your sample query; it's the same with and without the quotes.
$ gam print users query "name="TestUser1""
Getting all Users in G Suite account that match query (name=TestUser1) (may take some time on a large account)...
Got 0 Users:  - 
primaryEmail
$ gam print users query name=TestUser1  
Getting all Users in G Suite account that match query (name=TestUser1) (may take some time on a large account)...
Got 0 Users:  - 
primaryEmail

As soon as you really need quotes it breaks:
$ gam print users query "name="Test User1""

ERROR: User1 is not a valid argument for "gam print users"

This doesn't work because the API wants single quotes:
$ gam print users query name="Test User1" 
Getting all Users in G Suite account that match query (name=Test User1) (may take some time on a large account)...
Got 0 Users:  - 
primaryEmail

This doesn't work either because the single quotes get stripped.
$ gam print users query name='Test User1'
Getting all Users in G Suite account that match query (name=Test User1) (may take some time on a large account)...
Got 0 Users:  - 
primaryEmail

This is the valid way to express the query:
$ gam print users query "name='Test User1'"
Getting all Users in G Suite account that match query (name='Test User1') (may take some time on a large account)...
Got 2 Users: testuser1@rdschool.net - testuser1@rdschool.org
primaryEmail
testuser1@rdschool.net
testuser1@rdschool.org

I've thought about this and haven't thought of anything better than surrounding any argument with a space or comma in it with double quotes.

src/gam.py
```diff
@@ -5361,38 +5364,35 @@ def sendOrDropEmail(users, method='send'):
...
-    elif myarg == 'checkspam':
-      if method not in ['import']:
-        systemErrorExit(3, 'check_spam is only valid on importemail')
-      kwargs['neverMarkSpam'] = False
+    elif myarg == 'date':
+      msgHeaders['Date'] = getTimeOrDeltaFromNow(sys.argv[i+1])
+      if method in ['insert', 'import']:
+        kwargs['internalDateSource'] = 'dateHeader'
+      i += 2
+    elif method == 'import' and myarg == 'nevercheckspam':
```
jay0lee: API defaults to enabling check spam if this isn't set. That doesn't make a lot of sense to me, especially in the context of GAM here, if the admin is running the command to put the message in the user's mailbox, it almost certainly shouldn't go to Spam. I think I had this wrong but lets make neverMarkSpam=True the default for imports unless checkspam argument is supplied by user on GAM command line.

PR #962, opened 2019-07-01
Handle group members with no status

taers232c:


src/gam.py
```diff
@@ -9173,6 +9173,11 @@ def checkGroupExists(cd, group, i=0, count=0):
     entityUnknownWarning('Group', group, i, count)
     return None
 
+def _checkMemberRoleIsSuspended(member, validRoles, isSuspended):
+  memberStatus = member.get('status', 'UNKNOWN')
+  return ((not validRoles or member.get('role', ROLE_MEMBER) in validRoles) and
```
jay0lee: that's a pretty complex return. I'd rather see this broken into clearer logic across multiple lines.

PR #971, opened 2019-07-10
Convert team to shared in vault export; prepare for the future

taers232c:
Restore _getValidCourseStates(croom), it is used by _getCourseStates in print courses/course-participants

### DIFF
--- a/src/gam.py
+++ b/src/gam.py
@@ -1974,9 +1974,6 @@ def doDelCourse():
   callGAPI(croom.courses(), 'delete', id=courseId)
   print('Deleted Course %s' % courseId)
 
-def _getValidCourseStates(croom):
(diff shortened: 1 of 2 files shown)


src/gam.py
```diff
@@ -1994,7 +1997,7 @@ def getCourseAttribute(myarg, value, body, croom, function):
   elif myarg in ['owner', 'ownerid', 'teacher']:
     body['ownerId'] = normalizeEmailAddressOrUID(value)
   elif myarg in ['state', 'status']:
-    validStates = _getEnumValuesMinusUnspecified(croom._rootDesc['schemas']['Course']['properties']['courseState']['enum'])
+    validStates = _getValidCourseStates(croom)
```
jay0lee: I don' really see the point in calling 2 one line functions here, just call _getEnumValuesMinusUnspecified...

PR #974, opened 2019-07-16
Cleanup

taers232c:
parent is not valid with use project


src/gam.py
```diff
@@ -7894,9 +7893,10 @@ def convertGCPFolderNameToID(parent, crm2):
   # just need to be specific.
   body = {'pageSize': 1000, 'query': 'displayName="%s"' % parent}
   folders = callGAPI(crm2.folders(), 'search', body=body)
-  if not 'folders' in folders or len(folders['folders']) == 0:
+  folders = folders.get('folders', [])
```
jay0lee: we need to get out of the habit of "re-typing" a value like this. Lets just change it to:
```
   results = callGAPI(....
   folders = results.get('folders', [])
```

PR #980, opened 2019-07-26
I suggest match_users instead of if_users

taers232c:


src/gam.py
```diff
@@ -9714,8 +9714,8 @@ def doUpdateMobile():
       if body['action'] not in ['admin_remote_wipe', 'admin_account_wipe', 'approve', 'block', 'cancel_remote_wipe_then_activate', 'cancel_remote_wipe_then_block']:
         systemErrorExit(2, 'action must be one of wipe, wipeaccount, approve, block, cancel_remote_wipe_then_activate, cancel_remote_wipe_then_block; got %s' % body['action'])
       i += 2
-    elif myarg == 'ifusers':
-      only_users = getUsersToModify(entity_type=sys.argv[i+1].lower(), entity=sys.argv[i+2])
+    elif myarg == 'matchusers':
```
jay0lee: that's fine but lets keep ifusers as an (undocumented) alternative as I already have an admin using it.

PR #998, opened 2019-08-18
Handle more errors with short URL oauth create

taers232c:


src/gam.py
```diff
@@ -106,11 +106,11 @@ def wrapped_request_method(self, *args, **kwargs):
 google_auth_httplib2.AuthorizedHttp.request = _request_with_user_agent(
   google_auth_httplib2.AuthorizedHttp.request)
 
-def _createHttpObj(cache=None, override_min_tls=None, override_max_tls=None):
+def _createHttpObj(cache=None, timeout=None, override_min_tls=None, override_max_tls=None):
   tls_minimum_version = override_min_tls if override_min_tls else GC_Values[GC_TLS_MIN_VERSION]
   tls_maximum_version = override_max_tls if override_max_tls else GC_Values[GC_TLS_MAX_VERSION]
   return httplib2.Http(ca_certs=GC_Values[GC_CA_FILE], tls_maximum_version=tls_maximum_version, tls_minimum_version=tls_minimum_version,
-                       cache=cache)
+                       cache=cache, timeout=timeout)
```
jay0lee: Did you test if timeout=None has same behavior as it not being specified? I believe default is something like timeout=2minutes and we don't want to override that.

taers232c:
From `httplib2/__init__.py`
```
class Http(object):
    """An HTTP client that handles:                                                                                                                                   
                                                                                                                                                                      
    - all methods                                                                                                                                                     
    - caching                                                                                                                                                         
    - ETags                                                                                                                                                           
    - compression,                                                                                                                                                    
    - HTTPS                                                                                                                                                           
    - Basic                                                                                                                                                           
    - Digest                                                                                                                                                          
    - WSSE                                                                                                                                                            
                                                                                                                                                                      
    and more.                                                                                                                                                         
    """

    def __init__(
        self,
        cache=None,
        timeout=None,
        proxy_info=proxy_info_from_environment,
        ca_certs=None,
        disable_ssl_certificate_validation=False,
        tls_maximum_version=None,
        tls_minimum_version=None,
    ):
```

PR #999, opened 2019-08-21
handle empty recoveryPhone

taers232c:


jay0lee:
It wouldn't hurt to also add a user update comment to travis.yaml that handles clearing users recovery phone. That way it's routinely getting tested.

PR #1002, opened 2019-08-29
Add timeoffset option to gam version

taers232c:


jay0lee:
Service account is most impacted by wrong time so let's make it a default part of that check before we check each scope. If we're off by 30 or more seconds we should FAIL with an error to fix clock.

PR #1035, opened 2019-10-17
Fix multiprocessing imports, check for 32-bit GAM on 64-bit Windows

taers232c:


src/gam.py
```diff
@@ -52,8 +52,7 @@
 import webbrowser
 import zipfile
 import http.client as http_client
-from multiprocessing import Pool
-from multiprocessing import freeze_support
+from multiprocessing import Pool,freeze_support,set_start_method
```
jay0lee: I'm concerned generic names like these run the risk of colliding with something else down the road. Let's do:

    from multiprocessing import Pool as mp_pool
    from multiprocessing import freeze_support as mp_freeze_support
    from multiprocessing import set_start_method as mp_set_start_method

we could make that one long line but 3 lines above are preferable for readability.

PR #1047, opened 2019-12-03
Begin breaking apart gam.py into logical pieces

ejochman:
Start with one of the deepest parts of the stack, Google API request execution calls and associated errors. Critical information printing functions and application control logic are also broken out into their own components.

This change also adds unit tests for migrated content and makes code more PEP8 compliant.

This commit starts work on  jay0lee/GAM#147


ejochman:
@jay0lee I won't claim to be familiar with the build environment you're using, so if there's anywhere that these new files needed to be added to a build config, let me know.

daethnir:
@ejochman if you want to improve GAM, I'd start instead with the https://github.com/taers232c/GAMADV-XTD3 source. It is a better codebase (proper exit statuses, support for more features, python3 compatible) and tends to get new features earlier than GAM itself, while maintaining command line compatibility.

jay0lee:
This looks like a good start, for the unit tests you added, can we also add them to .travis.yaml so they are executed on all changes?

PR #1061, opened 2019-12-28
Allow forks of jay0lee/GAM to use Travis CI

ejochman:
Restrict e2e tests to only be run from jay0lee's repo, since they require certain secrets from that environment.

This enables forks to be setup with Travis to run all other applicable build tasks without having to modify the forked yaml.


ejochman:
Doesn't seem to be working as expected, so let me try this again.

PR #1064, opened 2019-12-31
When rotating an sakey, only delete the existing key, no others

taers232c:


jay0lee:
I think the key id is stored in service file or else can be calculated (sha1 sum of key? Cloud docs would say) so we can save on the list API call.

I'd prefer default be to wipe all previous user managed keys. Otherwise attacker who got old key could use it to generate a separate new key that never gets removed. Let's make your behavior another option, something like replace_only as opposed to leave_existing which doesn't remove anything.

taers232c:
What about?
````
gam create sakey - Make a new key, leave other keys
gam rotate sakey - Rotate existing key, leave other keys
gam replace sakeys - Generate new key, ldelete other keys
```

jay0lee:
There's rare use cases for having > 1 key and many paths to abuse or forget
about them if you do. It's a security risk. Keeping old keys around should
be a deliberate choice.

Jay

taers232c:
That's what I intended: rotate == keep old, replace == delete old
````
gam create sakey - Make a new key, leave other keys
gam rotate sakey - Rotate existing key, leave other keys
gam replace sakeys - Generate new key, delete other keys
```
I'm open to a different way of expressing the different modes.

jay0lee:
I want to keep anything that leaves old keys around a non-default option
that must be explicitly chosen by the user. Having separate commands for
each doesn't achieve that. Having handling of old keys as an argument does.

taers232c:
How about:
````
gam update sakeys - Generate new key, delete all existing keys
gam update sakeys retain_existing- Generate new key, retain existing keys
gam update sakeys replace_current - Generate a new key, delete current key, leave other keys
```

jay0lee:
Let's keep the rotate syntax. If your up for it in ready to make local generation the default also and leave Google generation as an option argument.

Issue #1063, opened 2019-12-30
Crash on credential refresh


jay0lee:
https://travis-ci.org/jay0lee/GAM/jobs/630968578#L1068

we seem to crash when we get a 502 response and attempt to refresh. request is a function (at least it is when we're using service account credentials). Need to figure out how to properly refresh credentials here.

jay0lee:
$gam users "$newbase-bulkuser-07 $newbase-bulkuser-08 $newbase-bulkuser-09" modify messages query in:anywhere maxtomodify 99999 addlabel IMPORTANT addlabel STARRED doit; fi
Searching messages for travis-test-7-1577718881679-bulkuser-07@pdl.jaylee.us
Got 3 messages for user travis-test-7-1577718881679-bulkuser-07@pdl.jaylee.us
modify 3 messages
Traceback (most recent call last):
  File "/home/travis/build/jay0lee/GAM/src/gapi/__init__.py", line 84, in call
    return method(**parameters).execute()
  File "/home/travis/virtualenv/python3.7.5/lib/python3.7/site-packages/googleapiclient/_helpers.py", line 130, in positional_wrapper
    return wrapped(*args, **kwargs)
  File "/home/travis/virtualenv/python3.7.5/lib/python3.7/site-packages/googleapiclient/http.py", line 856, in execute
    raise HttpError(resp, content, uri=self.uri)
googleapiclient.errors.HttpError: <HttpError 502 when requesting https://www.googleapis.com/gmail/v1/users/me/messages/batchModify?prettyPrint=false returned "Bad Gateway">
During handling of the above exception, another exception occurred:
Traceback (most recent call last):
  File "gam.py", line 14689, in <module>
    sys.exit(ProcessGAMCommand(sys.argv))
  File "gam.py", line 14463, in ProcessGAMCommand
    doProcessMessagesOrThreads(users, 'modify', 'messages')
  File "gam.py", line 5853, in doProcessMessagesOrThreads
    userId='me', **kwargs)
  File "/home/travis/build/jay0lee/GAM/src/gapi/__init__.py", line 94, in call
    service._http.request.credentials.refresh(create_http())
AttributeError: 'function' object has no attribute 'credentials'

taers232c:
I'll submit a PR for this when PR #1064 is done.

taers232c:
Can you reproduce this? How does #1066 affect it?

jay0lee:
[...]

PR #1066, opened 2020-01-01
Cleanup

taers232c:
gam,py:
In show sakays, only mark USER-MANAGED keys as current: true/false

gapi/__init__.py:
pylint cleanup: import order, indentation

gapi/errors.py:
pylint cleanup: import order, unused import, indentation

Recognize 502 Bad Gateway/Gateway Timeout: treat as retryable errors

This last change does not directly handle the refresh problem in Issue #1063 but in my version this seems to be the right solution to the 502 gateway errors

### DIFF
--- a/src/gam.py
+++ b/src/gam.py
@@ -7840,7 +7840,7 @@ def _formatOAuth2ServiceData(private_key, private_key_id):
 
 def doShowServiceAccountKeys():
   iam = buildGAPIServiceObject('iam', None)
-  keyTypes = 'USER_MANAGED'
+  keyTypes = None
   i = 3
   while i < len(sys.argv):
     myarg = sys.argv[i].lower().replace('_', '')
@@ -7867,7 +7867,7 @@ def doShowServiceAccountKeys():
     print('{0}: {1}'.format(parts[i][:-1], parts[i+1]))
   for key in keys:
     key['name'] = key['name'].rsplit('/', 1)[-1]
-    if key.get('keyType', '') == 'USER_MANAGED' and key['name'] == currentPrivateKeyId:
+    if key['name'] == currentPrivateKeyId:
       key['usedToAuthenticateThisRequest'] = True
   print_json(None, keys)
 
--- a/src/gam.py
+++ b/src/gam.py
@@ -7867,7 +7867,8 @@ def doShowServiceAccountKeys():
     print('{0}: {1}'.format(parts[i][:-1], parts[i+1]))
   for key in keys:
     key['name'] = key['name'].rsplit('/', 1)[-1]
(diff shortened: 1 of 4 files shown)


src/gam.py
```diff
@@ -7867,7 +7867,8 @@ def doShowServiceAccountKeys():
     print('{0}: {1}'.format(parts[i][:-1], parts[i+1]))
   for key in keys:
     key['name'] = key['name'].rsplit('/', 1)[-1]
-    key['current'] = key['name'] == currentPrivateKeyId
+    if key.get('keyType', '') == 'USER_MANAGED':
```
jay0lee: I'd prefer to just set current=true on the one key and leave it unset instead of false for everything else. Also current is a bit ambiguous, we should make it a bit more clear that it's the key GAM is currently using. Maybe, used_to_authenticate_this_request ?

Additionally, I'm not opposed to a command that will revoke a single key along the lines of:

gam delete sakey <id>

what I don't want is commands like "gam create sakey" that will make it to easy to end up with a whole bunch of keys that are a challenge for an admin to identify and manage, that just weakens our security story while adding no real value (admins shouldn't need to do that).

taers232c:
This command already exists:
gam delete sakey|sakeys <ServiceAccountKeyList>+ [doit]

src/gam.py
```diff
@@ -7867,7 +7867,8 @@ def doShowServiceAccountKeys():
     print('{0}: {1}'.format(parts[i][:-1], parts[i+1]))
   for key in keys:
     key['name'] = key['name'].rsplit('/', 1)[-1]
-    key['current'] = key['name'] == currentPrivateKeyId
+    if key.get('keyType', '') == 'USER_MANAGED' and key['name'] == currentPrivateKeyId:
```
jay0lee: no reason to check if it's user managed here and we may use GOOGLE_MANAGED in the future, thus no reason to do this check. Also, let's make showing all keys the default.

src/gapi/errors.py
```diff
@@ -1,12 +1,12 @@
 """GAPI and OAuth Token related errors methods."""
 
+from enum import Enum
 import json
+#import googleapiclient.errors
```
jay0lee: this is breaking unittests:

https://travis-ci.org/jay0lee/GAM/jobs/631580165#L407

jay0lee: if it's unused then the better solution would be to fix the unittests. @ejochman any ideas here?

jay0lee: for now I've modiified the tests so they should pass without the unused import here. Let's go ahead and remove it for now and @ejochman can let us know if this is voiding the tests.

PR #1071, opened 2020-01-06
Move transport customizations to their own module

ejochman:
This helps to accomplish a few things:
1) Makes the forced user-agent customization on HTTP requests a bit
clearer by subclassing the targeted objects (as opposed to hiding the
behavior behind a forced override of the google_auth_httplib2 object
methods)
2) Standardizes the creation of HTTP objects. These objects can still
largely be customized, but using a single creation mechanism will
standardize a default and streamline creation, thereby decreasing the
code that would otherwise be replicated in the caller
3) Moves create_http() to a more general purpose module, since it will
likely be used by more than gapi-related methods.

This change continues work on jay0lee/GAM#147

### DIFF
--- a/src/transport_test.py
+++ b/src/transport_test.py
@@ -28,12 +28,12 @@ class CreateHttpTest(unittest.TestCase):
     self.assertEqual(http.ca_certs, transport.GC_Values[transport.GC_CA_FILE])
 
   def test_create_http_sets_tls_min_version(self):
-    http = transport.create_http(override_min_tls=1111)
-    self.assertEqual(http.tls_minimum_version, 1111)
+    http = transport.create_http(override_min_tls='TLSv1_1')
+    self.assertEqual(http.tls_minimum_version, 'TLSv1_1')
 
   def test_create_http_sets_tls_max_version(self):
-    http = transport.create_http(override_max_tls=9999)
-    self.assertEqual(http.tls_maximum_version, 9999)
+    http = transport.create_http(override_max_tls='TLSv1_3')
+    self.assertEqual(http.tls_maximum_version, 'TLSv1_3')
 
   def test_create_http_sets_cache(self):
     fake_cache = {}
--- a/src/transport_test.py
+++ b/src/transport_test.py
@@ -1,0 +1,179 @@
+"""Tests for transport."""
+
+import unittest
+from unittest.mock import MagicMock
+from unittest.mock import patch
+
+from gam import SetGlobalVariables
+import google_auth_httplib2
+import httplib2
+
+import transport
+
+
+class CreateHttpTest(unittest.TestCase):
+
+  def setUp(self):
+    SetGlobalVariables()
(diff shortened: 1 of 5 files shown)


src/transport_test.py
```diff
@@ -0,0 +1,179 @@
...
+    http = transport.create_http()
+    self.assertIsNone(http.cache)
+    self.assertIsNone(http.timeout)
+    self.assertEqual(http.tls_minimum_version,
+                     transport.GC_Values[transport.GC_TLS_MIN_VERSION])
+    self.assertEqual(http.tls_maximum_version,
+                     transport.GC_Values[transport.GC_TLS_MAX_VERSION])
+    self.assertEqual(http.ca_certs, transport.GC_Values[transport.GC_CA_FILE])
+
+  def test_create_http_sets_tls_min_version(self):
```
jay0lee: should we be testing with actual TLS version from:

https://docs.python.org/3/library/ssl.html#ssl.TLSVersion

here?

ejochman: We could, sure. I'm just validating the behavior that already existed.

What was the original purpose of saving these values as globals, rather than pulling them from the library directly, when used? Is GAM allowing users to set their own min/max through an environment var? I just want to avoid a case where the tests fail because the min/max values in the globals are different than that of the SSL lib.

ejochman: @jay0lee friendly ping. Any preference, here, on how this should behave in regards to the TLS lib defaults vs. the saved GAM globals?

jay0lee: Google servers support TLS 1.2+ as should GitHub and the occassional other HTTPS server GAM talks to (few commands like set user photo can pull from an HTTP server). However there are two cases where admin may need something less than TLS 1.2:
- Some proxies need to force a certain older TLS version on the client (bad I know but it happens).
- some commands will pull from custom web server that may need older TLS.
GAM should default to requiring 1.2 or higher. It's possible one of above legit scenarios may require a downgrade and an environment variable can do that.

So all that to say GAM should probably throw an error when the value isn't one of the ones in `TLSVersion`. That change can come later but for now let's make sure the tests are using legit versions.

ejochman: OK. I think I misunderstood your original post on this. I get it now. Looks like the `ssl.TLSVersion` enum is only available in Python 3.7+, though, so using it here would cause some compatibility issues with 3.5 and 3.6 which I believe you're still testing.

`TLSVersion` is an [IntEnum](https://docs.python.org/3/library/enum.html#intenum), but I see you're [currently defaulting the min version to a string `'TLSv1_2'`](https://github.com/jay0lee/GAM/blob/master/src/var.py#L853). I'd be curious to know if that really works. Because, from the docs, it doesn't seem like it should. But, I have no other proof to back it up. For now, I could update the tests to use values like `'TLSv1_1'` and `'TLSv1_3'` to match what's there already.

PR #1075, opened 2020-01-16
Eliminate adminCreated as a writable field in update group

taers232c:


jay0lee:
this should be an undocumented but still settable feature. Add a comment to the actual setting code to note that it doesn't actually modify the setting and remove from the commands txt file.

PR #1091, opened 2020-02-09
f string formatting, sixth of many

taers232c:
Got message cleanup


src/gapi/__init__.py
```diff
@@ -16,6 +16,16 @@
                  MESSAGE_API_ACCESS_DENIED, MESSAGE_SERVICE_NOT_APPLICABLE)
 
 
+TOTAL_ITEMS_MARKER = '%%total_items%%'
```
ejochman: @taers232c When adding new content to these new files, can we please add docstrings, stick with PEP8 style ([method names are lowercase_with_underscores](https://www.python.org/dev/peps/pep-0008/#method-names-and-instance-variables)), and add tests for new functionality? Otherwise, I fear this codebase will eventually going to decay back to its original state, which makes it much harder to introduce novel changes.

Issue #1120, opened 2020-03-19
gapi.calendar.buildCalendarDataGAPIObject is making an unnecessary credential request before falling back to admin credentials


ejochman:
`gapi.calendar.buildCalendarDataGAPIObject` currently makes an [unnecessary credentials request](https://github.com/jay0lee/GAM/blob/d78d68b4da5c64527fe74c233cdb0f7d69305e3b/src/gapi/calendar.py#L35), when trying to build a calendar service stub for a resource calendar operation. It tries to pass the resource calendar email address as the act_as parameter. This request will always generate a `RefreshError` because a resource calendar cannot be impersonated as an account. It relies on an underlying behavior of `gapi.handle_oauth_token_error` to return None, which will in turn cause `gapi.calendar.buildCalendarDataGAPIObject` to request admin credentials instead.

As a stop-gap solution, `gapi.calendar.buildCalendarDataGAPIObject` should look for a resource calendar email address and request admin credentials in that case, before trying to impersonate the given email address. Long-term, this functionality should be better architected, but will require some major refactoring in credential operations.

PR #1122, opened 2020-03-20
Avoid requests to impersonate a resource calendar

ejochman:
Fixes jay0lee/GAM#1120


taers232c:
Isn't the same thing true for groups calendars?
```
if ((not calname.endswith('@resource.calendar.google.com') and
    (not calname.endswith('@groups.calendar.google.com')):
```

PR #1123, opened 2020-03-21
Another non-user calendar

taers232c:


jay0lee:
Let's simplify to ends with calendar.google.com.

PR #1126, opened 2020-03-23
Centralize OAuth2.0 Credential logic

ejochman:
Adds a Credentials class that centralizes and handles most existing logic related to OAuth2.0 credentials, including generation, storage, file locking, and attribute retrieval. This is a step towards minimizing the duplicated code that handles credentials in various methods and hiding the nitty-gritty behind convenient accessors. The goal is to eventually get to a point where there are 2 credential entry points: `auth.get_admin_credentials()` and `auth.get_credentials_for_user(user)`. Then, we can slowly move toward using service account / impersonated credentials for all operations and scrap the need for user consented credentials all together.


ejochman:
While I'm pretty certain I've covered all corner cases that the previous code had, I don't have any really old installations to test with, so I'd appreciate if someone can proof that old credentials get un-pickled and updated correctly. Worst case, someone would have to recreate their credentials file.

jay0lee:
This is excellent but I see one issue. We can't move to 100% service account credentials unless Google adds the ability to limit which users DwD works against, currently it can be *any* user. Consider the following scenario:

Central IT wants to allow tier 2 helpdesk personnel (t2-hd@acme.com) the ability to reset end user passwords with GAM. Today that's really easy to achieve, setup t2-hd@acme.com as a delegated admin with rights to reset user passwords. This lets them reset regular user passwords but they can't elevate their privileges by resetting the passwords of super admins or other delegated admins. For the GAM install, ignore service accounts entirely and just get an oauth2.txt token for t2-hd@acme.com scoped to allow only user updates.

Now consider the same scenario with service account DwD, it can't be done. The oauth2service.json service account would be authorized to use the scope `https://www.googleapis.com/auth/admin.directory.user` as ANY user in the domain. The helpdesk person could choose to be t2-hd@acme.com but they can also act as super-admin@acme.com and gain super admin rights by resetting super admin passwords.

I recognize two OAuth flows adds significant complexity but I'm not willing to lose GAM's current ability to strictly scope and limit permissions in order to simplify here.

@ejochman in light of this, do you want to think about a different approach with this PR or does it still work? Again, I really appreciate the work you are doing here.

Jay

ejochman:
Ah. Ok, that's a case I hadn't considered -- I didn't know folks used the tool this way. Let's keep moving forward here and I'll take that into consideration as I incorporate more changes. Maybe we get to a point where, during GAM setup, we can have the user choose between service account and consented credentials, explaining the cost/benefit at the same time. Hopefully, once we can abstract the credential logic from the rest of the business logic, then it's just a single conditional deciding whether we attempt to build oauth2.0 or service account credentials for the action being requested, and callers don't need to decide for themselves.

PR #1132, opened 2020-03-26
Fix gam-install.sh

taers232c:


jay0lee:
Thanks, already commited it.

PR #1133, opened 2020-03-26
Fix oauth create

taers232c:


jay0lee:
https://github.com/jay0lee/GAM/blob/master/src/gam.py#L10424

any idea why we;re returning a set to begin with above?

taers232c:
It would appear to be to avoid having the same scope in the list twice.
--
ross.scroggs@gmail.com

jay0lee:
So maybe it should be a:

list(set(selected_scopes))

your fix is fine for now but we should get the function to return a list as
expected ultimately.

Jay Lee

PR #1149, opened 2020-04-04
Credentials must be current to get token values

taers232c:
```
$ gam calendar testuser1@domain.com printevents
  File "/Users/Ross/GAM/gam.py", line 11603, in <module>
    sys.exit(ProcessGAMCommand(sys.argv))
  File "/Users/Ross/GAM/gam.py", line 11203, in ProcessGAMCommand
    gapi.calendar.printEvents()
  File "/Users/Ross/GAM/gapi/calendar.py", line 167, in printEvents
    calendarId, cal = buildCalendarDataGAPIObject(sys.argv[2])
  File "/Users/Ross/GAM/gapi/calendar.py", line 32, in buildCalendarDataGAPIObject
    calendarId = normalizeCalendarId(calname)
  File "/Users/Ross/GAM/gapi/calendar.py", line 20, in normalizeCalendarId
    GC_Values[GC_DOMAIN] = __main__._getValueFromOAuth('hd')
  File "/Users/Ross/GAM/gam.py", line 7822, in _getValueFromOAuth
    return google.oauth2.id_token.verify_oauth2_token(credentials.id_token, request)[field]
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/site-packages/google/oauth2/id_token.py", line 140, in verify_oauth2_token
    return verify_token(
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/site-packages/google/oauth2/id_token.py", line 123, in verify_token
    return jwt.decode(id_token, certs=certs, audience=audience)
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/site-packages/google/auth/jwt.py", line 239, in decode
    _verify_iat_and_exp(payload)
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/site-packages/google/auth/jwt.py", line 190, in _verify_iat_and_exp
    raise ValueError("Token expired, {} < {}".format(latest, now))
ValueError: Token expired, 1585928134 < 1586010953
```

### DIFF
--- a/src/gam.py
+++ b/src/gam.py
@@ -7817,6 +7817,9 @@ def doCreateResoldCustomer():
 def _getValueFromOAuth(field, credentials=None):
   if not credentials:
     credentials = auth.get_admin_credentials()
+  if credentials.expired:
+    request = transport.create_request()
+    credentials.refresh(request)
   return credentials.get_token_value(field)
 
 def doGetMemberInfo():


src/gam.py
```diff
@@ -7817,6 +7817,9 @@ def doCreateResoldCustomer():
 def _getValueFromOAuth(field, credentials=None):
   if not credentials:
     credentials = auth.get_admin_credentials()
+  if credentials.expired:
```
ejochman, line 7820: This should really be handled in Credentials#get_token_value, itself. Otherwise, other callers are going to experience the same thing. I can draw up a PR with the fix

jay0lee:
I just moved the logic into get_admin_credentials:

https://github.com/jay0lee/GAM/commit/c83430a537531399dfdd1fb30e7357920c396a49

Jay Lee

ejochman:
c83430a537531399dfdd1fb30e7357920c396a49 doesn't address a latent expiration case like :

Suppose a set of credentials stored on disk expires at t100. The token data for those credentials has never been fetched, and therefore never written to disk.

1. At t99, credentials are loaded from disk but are not yet expired, thus they are not refreshed.
2. The caller does some other, unrelated operations, then needs to fetch a value from the token data.
3. At t101, `Credentials#get_token_value` is called, but fails with the error seen in the OP, as the token is now expired

I've created https://github.com/jay0lee/GAM/pull/1152 to fix this by refreshing credentials just before attempting to fetch the token data. Note that I also removed the refresh added to `auth#get_admin_credentials` since it should be unnecessary. `AuthorizedHttp` will handle credential refreshes before a request is made, if it's needed.  This is one weird corner case where an attribute of the the credentials is being used as a payload in the request, rather than being used as auth for the request (where `AuthorizedHttp` would know to refresh).

PR #1163, opened 2020-04-21
Fix URL to avoid SSL issue

taers232c:


jay0lee:
What SSL issue?

taers232c:
I had a user that did this: gam update verify domain.com txt
and got: ERROR: A failure in the SSL library occurred (_ssl.c:1108)

When we updated th URL to have .com, the problem went away,
--
ross.scroggs@gmail.com

jay0lee:
That's likely an issue in their network setup then (possibly doing SSL inspection). Unless this is easily reproducible I'd rather leave this as is so that such environment issues are exposed.

taers232c:
OK. dns.google and dns.google.com wind up at the same service, what's the advantage to using dns.google?
--
ross.scroggs@gmail.com

jay0lee:
dns.google.com is being turned down. See:

https://developers.google.com/speed/public-dns/docs/doh/migration?hl=en

changing to dns.google.com now will just result in a 302 redirect back to
dns.google in June...

Jay Lee

taers232c:
Got It. Thanks
--
ross.scroggs@gmail.com

PR #1164, opened 2020-04-21
Keep those Mac OS zsh people happy/Fix documentation

taers232c:
Make a .zshrc and set alias


jnance86:
As a Mac OS zsh user i am not happy (install not working)

Issue #720, opened 2018-04-16
Make GAM pip installable


jay0lee:
It should be possible to install GAM via pip. This does not replace the binary installation scripts but does offer a simple install for users who already have Python 2.7 and pip on their system.

First step is to complete #719 but once that's done the process should be pretty straightforward:

https://packaging.python.org/tutorials/distributing-packages/

LewisLebentz:
Would be super useful for FaaS, unfortunately I don't have the knowledge to create a PR to help.

It's the one thing that would fully unlock GAMs potential for us, helping with automation and delegating out controls that aren't possible via admin roles.

Issue #874, opened 2019-04-02
Codebase needs a style guide


ejochman:
There are a couple issues open right now that suggest a general restructuring of GAM code, which I agree would be good. However, one simple fix that would help both with code structure and readability of contributed code is a style guide that could be enforced during PR review. 

The main file is over 13k lines, at this point, and lacks a consistent style and structure throughout, making it difficult for new contributors to even find their footing. There are several existing published style guides that could be utilized. I'm a bit biased, but I believe [Google's published Python style guide](http://google.github.io/styleguide/pyguide.html) is generally popular amongst different organizations. That said, I'd be happy with just about any choice at this point, as long as it improves the consistency.

Some examples of readability issues in current code:
- Lack of docstrings on most methods
- camelCase and snake_case inconsistency across variables and method declarations
- Really long character line limits causing horizontal scroll to be necessary, even on large displays
- Line break indentation inconsistency

ejochman:
Reopening as a lingering issue

PR #1165, opened 2020-04-22
Refactor into Python package format

### LINKED ISSUE #23: Refactor GAM to be "import"able
I'd like to use some of the GAM logic in separate scripts. GAM already does a lot of the hard work to set up API objects and the like.

I'd like to refactor the code to start making it reusable in new scripts.

My first step would be to pull out the object initialization into a new file, such as gamlib.py, which would be used by gam.py. Everything that ends up in gamlib.py would be something that is built to be reused - all OO, no globals, etc.

Would you like to discuss this and tell me what requirements you'd have for it to be something you're willing to merge in?

ejochman:
-Updates import statements to be absolute vs implicitly relative
-Uses import syntax that minimizes the need to update references in code
and/or reformat affected lines (e.g. `import gapi.directory` becomes `from gam.gapi import directory as
gapi_directory`)
-Adds a `__main__.py` such that the package can be executed on its own using standard `python3 -m gam` syntax
-Replaces `__main__` import hack with a package import
-Updates the GAM path to be the package's parent dir

-Provides backwards compatibility for clients that execute gam.py file directly

Fixes #23
Makes progress on #720


taers232c:
Have you run pylint against all of the modules?

Ross

ejochman:
Is there a published pylint configuration we're using? I'm wholly in support of making sure each PR is linted, but we should have an agreed upon config first, and then setup a github check for PRs -- just so we're all consistent. See #874

taers232c:
You/Jay and I have different styles, so you guys can adopt whatever style you want.
I'm mostly interested in semantic errors: missing imports, uninitialized variables, unknown modules, ...

jay0lee:
I am onboard with adopting Google's style guide as discussed in #874 . I'll acccept a PR to add that to the docs where necessary.

ejochman:
@jay0lee @taers232c Cool. I'll put together a PR for a pre-commit

taers232c:
I'm not sure how this fits into a PR. I run pylint before I submit code (OK, in my advanced age I occasionally forget) to keep semantic errors out of submitted code. I see this as setting standard operating procedure, pylint the code and then submit it.

ejochman:
@taers232c We can add a [pre-commit](https://pre-commit.com/#intro) which will enforce that certain checks pass before a commit is made. One of those can be running [yapf](https://github.com/google/yapf) with `--style=google`. yapf will auto-correct the issues it finds that violate the style guide. That should include everything that pylint would flag and more.

taers232c:
Look at PR #1134, scroll down to utils.c
This is not a style issue, it's a bug (getDeltaDate should be get_delta_date); it should not have been submitted. I use pylint to find these errors before submission.
```
      argstr = (datetime.datetime(today.year, today.month, today.day)+getDeltaDate(argstr)).strftime(YYYYMMDD_FORMAT)
      argstr = (datetime.datetime(today.year, today.month, today.day)+get_delta_date(argstr)).strftime(YYYYMMDD_FORMAT)
```

ejochman:
@taers232c Agreed. I updated my earlier comment. I'm going to use yapf to format, but I'll also add a pre-commit to run pylint which will flag these errors before the commit.

ejochman:
#1166 will add the linting/formatting. I manually fixed pylint errors in this PR. I would suggest we pull this first before pulling #1166 to make the commit history a bit easier to read. Otherwise, we'll have some large file renaming, in addition to large scale refactoring to correct pylint issues in the same PR. For history sake, I think we should first re-organize, then start fresh on linting/formatting, going forward.

I think this PR is ready for review now and will open it up. @jay0lee I could use your help in figuring out why Travis is failing with missing args to a_atleast_b.py  in the library version comparisons on Linux/MacOS builds.

jay0lee:
Thanks @ejochman , this is awesome! Regarding the issues with a_atlesat_b.py, it looks like everything is working for the "source" travis test builds where we just run GAM using Python 3.6, 3.7 and daily but it's failing for all the build VMs that use PyInstaller. PyInstaller succeeds in building a GAM executable but running it never produces any output or does anything.

Few issues I see here:
- [gam.spec](https://github.com/jay0lee/GAM/blob/master/src/gam.spec) needs to be updated to target gam/__main__.py as the main entry point rather than gam.py. Seems like gam.py should work based on your backward compat changes but it doesn't and the GAM executables shouldn't depend on that anyway. Changing gam.py to gam/__main__.py in the gam.spec solved this for me.
- Currently PyInstaller puts the compiled GAM executables in a gam/ folder so they are ready for .zip files, etc. This no longer works since gam/ is now the source directory so we need to fix those build scripts. It's going to mean multiple per-OS changes to the building / packaging steps but it makes sense to fix. I'll see what I can figure out.

jay0lee:
For the 2nd point, we need to remove the ```--distpath=gam``` argument in each ```travis/<platform>-install.sh``` file so that PyInstaller defaults to creating binaries in a dist folder. Then we can figure out how to fix the tests and packaging steps.

jay0lee:
Here's a patch that should fix the Travis builds at least for the PR travis runs. The trick here is that when running against PRs, Travis doesn't upload the GAM binaries anywhere, it only does that on actual commits. So we may need to just commit this change and then inspect the resultant binary packages for any lingering issues.

[dist-cleanup.diff.txt](https://github.com/jay0lee/GAM/files/4523163/dist-cleanup.diff.txt)

ejochman:
@jay0lee with https://github.com/jay0lee/GAM/commit/18bc4598500c3f4c5b08cfaa0d5e0728269ed194 removing the OS-specific specs, do we need to update `linux-build.sh` and `macos-build.sh` to refer to `gam.spec`? It looks like they still point at the OS-specific files.

taers232c:
Already done: https://github.com/jay0lee/GAM/commit/18bc4598500c3f4c5b08cfaa0d5e0728269ed194

ejochman:
@taers232c Did you mean to link a different commit? I'm still seeing the old references at HEAD [here](https://github.com/jay0lee/GAM/blob/5f6306911fcdfee608e76a0c514ce612f44259a0/src/linux-build.sh#L7) It looks like the `-install.sh` files were updated, but not `-build.sh`

jay0lee:
Looks like I missed -build.sh when unifying the spec files, yes it should
be updated.

Jay Lee

taers232c:
You're right, I need to clean my glasses.

jay0lee:
Chatted offline with Erik and we'll take a look at the -build.sh files later, they're already broken anyway.

@taers232c I honestly haven't used the -build.sh files since standing up Travis, maybe we should just remove them?

taers232c:
I'd completely forgotten about them. The users of mine that have the source are just running the source as far as I know, they're not building their own executable version.
I'd toss them.
--
ross.scroggs@gmail.com

PR #1166, opened 2020-04-22
Add pre-commit config for formatting and linting

ejochman:
Adds a pre-commit config that runs several fixers, including YAPF with
"google" style and pylint for static analysis

Note: This is a local pre-commit only. Running a GitHub action to perform this action on PRs may be configurable.

Makes progress on #874 to define a style guide. May still need to setup a GitHub action to enforce on PRs.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,27 @@
+# See https://pre-commit.com for more information
+# See https://pre-commit.com/hooks.html for more hooks
+default_language_version:
+    python: python3.7
+
+repos:
+-   repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v2.4.0
+    hooks:
+    -   id: trailing-whitespace
+    -   id: end-of-file-fixer
+    -   id: check-yaml
+    -   id: check-docstring-first
+    -   id: name-tests-test
+    -   id: requirements-txt-fixer
+
+-   repo: https://github.com/pre-commit/mirrors-yapf
+    rev: v0.29.0
+    hooks:
+    -   id: yapf
+        args: [--style=google, --in-place]
+
+-   repo: https://github.com/PyCQA/pylint
+    rev: pylint-2.4.4
+    hooks:
+    -   id: pylint
+        args: [--output-format=colorized]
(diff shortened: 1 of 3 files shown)


jay0lee:
requirements.txt should only be what's necessary to build GAM, not develop. We may need some kind of dev requirements....

ejochman:
I moved the `pre-commit`requirement to `requirements-dev.txt` which appears to be a commonly utilized pattern, per https://pypi.org/project/requirements-dev.txt/

jay0lee:
Thanks @ejochman 

If I understand correctly we won't need Travis to install ```requirements-dev.txt``` but GAM developers should? What happens if pre-commit isn't installed? Will git fail or will it just allow the commit without running pre-commit?

ejochman:
I think developers would have to install `requirements-dev.txt` and then run `pre-commit install` to install pre-commit in their git hooks. From there, it should run before every `git commit`. It looks like there's a [github action](https://github.com/pre-commit/action) we could use to run it, but it doesn't seem to scope the pre-commit results to only the files modified in the commit, out of the box (there may be a way to do this, but I haven't read into the Actions api enough to figure it out). Running pre-commit on all files is great for the long-term where everything in the repo is clean and only new changes would introduce issues. However, right now, it's a bit of a mess, and each PR is going to have a laundry list of issues until we do a massive cleanup. See [this example](https://github.com/ejochman/GAM/runs/616207468?check_suite_focus=true). Contrast that with the local pre-commit where it will only run the linter/formatter against files that are in the commit, which is a bit easier to handle in most cases.

ejochman:
Sorry. To more specifically answer your question:

> Will git fail or will it just allow the commit without running pre-commit?

If the dev doesn't have pre-commit installed, nothing happens and the commit succeeds as it does today. If it is installed, then it will prevent them from committing until the issues are fixed and staged for commit, or they bypass the pre-commit with a flag.

My suggestion: We introduce this pre-commit config locally, first (just to get us all on a baseline config that we can work against). I can put some effort into auto-cleaning most of the codebase and followup with some easy linter fixes. Then, we can see where we're at and how much further we'd have to get, in order please all the linter/formatter errors and decide if it's the right time to add the github action. Thoughts?

PR #1177, opened 2020-04-30
Yet another OAuth2 token error

taers232c:
What exactly are the circumstances that cause this?


src/gam/gapi/errors.py
```diff
@@ -31,7 +31,6 @@ class GapiConditionNotMetError(Exception):
 class GapiCyclicMembershipsNotAllowedError(Exception):
     pass
 
-
```
jay0lee, line 34: Two spaces between classes I believe...

PR #1179, opened 2020-05-01
Update report documentation/fix gam create project

taers232c:


taers232c:
I've seen it take 25-45 seconds, at the moment it's quicker.

PR #1191, opened 2020-05-14
Update gam report usageparameters to account for Google API change

taers232c:


jay0lee:
Can we reuse the logic in ```_check_full_data_available(fullDataRequired='all')``` here?

taers232c:
I'll check it out

src/gam/gapi/reports.py
```diff
@@ -78,36 +76,29 @@ def showUsageParameters():
...
-                               one_day).strftime(YYYYMMDD_FORMAT)
+            result = gapi.call(endpoint,
+                               'get',
+                               throw_reasons=throw_reasons,
+                               date=tryDate,
+                               customerId=customerId,
+                               fields='warnings,usageReports',
+                               **kwargs)
+            warnings = result.get('warnings', [])
+            fullDataRequired = ['all']
```
jay0lee, line 87: this can be moved out of the ```while True``` loop. Otherwise looks good. Thanks.

jay0lee:
There's a chance that: a new parameter is added and starts showing on reports on a recent day that still only has partial results. So our loop should always add any parameters we get to all_parameters even if the results are partial. Then we keep adding any found parameters from recent days but we ensure our list of all_parameters is complete by looping until we find a date in the past which has no warnings / partial results.

PR #1236, opened 2020-09-11
Fix error; DASA suggestions

taers232c:


src/gam/auth/__init__.py
```diff
@@ -2,28 +2,30 @@
...
 
 
 def get_admin_credentials_filename():
     """Gets the name of the file that stores the admin account credentials."""
     # If the environment globals are loaded, use the set global value. It may have
     # some custom name in it. Otherwise, just use the default name.
-    if GC_Values[GC_OAUTH2_TXT]:
-        return GC_Values[GC_OAUTH2_TXT]
-    return DEFAULT_OAUTH_STORAGE_FILE
+    if not GC_Values[GC_ENABLE_DASA]:
```
jay0lee, line 25: take not out of the if statement and then swap the two return statements

src/gam/auth/__init__.py
```diff
@@ -33,9 +35,11 @@ def get_admin_credentials(api=None):
         raise oauth.InvalidCredentialsFileError
     with open(credential_file, 'r') as f:
         creds_data = json.load(f)
-    if 'token' in creds_data:
+    if not GC_Values[GC_ENABLE_DASA] and 'token' in creds_data:
```
jay0lee, line 38: same

PR #1237, opened 2020-09-12
Complete adminrole commands; add signout and turnoff2sv

taers232c:


src/gam/__init__.py
```diff
@@ -11203,7 +11203,7 @@ def ProcessGAMCommand(args):
             elif argument in ['gcpfolder']:
                 createGCPFolder()
             elif argument in ['adminrole']:
-                gapi_directory_roles.create()
+                gapi_directory_roles.createUpdate(False)
```
jay0lee, line 11206: Let's keep separate create and update functions even if they are simple wrappers to the main function.

taers232c:
I think that this is what you wanted.

src/gam/__init__.py
```diff
@@ -6186,6 +6186,34 @@ def getVacation(users):
                     print('None')
 
 
+def doSignout(users):
```
jay0lee, line 6189: move these into gapi/directory/users.py where all the other user stuff will eventually land. Names should be signout() and turn_off_2sv().

PR #1238, opened 2020-09-15
Enforce DASA requirements; define a DASA user

taers232c:


src/gam/__init__.py
```diff
@@ -555,6 +556,23 @@ def _getCfgRowFilter(itemName):
     if not GC_Values[GC_NO_UPDATE_CHECK]:
         doGAMCheckForUpdates()
 
+# domain must be set and customer_id must be set and != my_customer when enable_dasa = true
+    if GC_Values[GC_ENABLE_DASA]:
+        if not GC_Values[GC_DOMAIN]:
+            controlflow.system_error_exit(
+                3,
+                f'Environment variable GA_DOMAIN must be set when enabledasa.txt is present'
```
jay0lee, line 564: can we show the full path to enabledasa.txt here to help the user know what file to delete if needed?

src/gam/__init__.py
```diff
@@ -555,6 +556,23 @@ def _getCfgRowFilter(itemName):
...
+                )
+        if not GC_Values[GC_CUSTOMER_ID] or GC_Values[GC_CUSTOMER_ID] == MY_CUSTOMER:
+            controlflow.system_error_exit(
+                3,
+                f'Environment variable CUSTOMER_ID must be set and not equal to my_customer when enabledasa.txt is present'
```
jay0lee, line 569: add "Your customer ID can be found at admin.google.com > Account > Profile."

src/gam/__init__.py
```diff
@@ -535,6 +535,7 @@ def _getCfgRowFilter(itemName):
                       fileAbsentValue=True)
     _getOldSignalFile(GC_NO_SHORT_URLS, 'noshorturls.txt')
     _getOldSignalFile(GC_NO_UPDATE_CHECK, 'noupdatecheck.txt')
+    _getOldEnvVar(GC_DASA_ADMIN, 'DASA_ADMIN')
```
jay0lee, line 538: I don't understand what this value is for?

src/gam/__init__.py
```diff
@@ -8493,6 +8511,12 @@ def doCreateResoldCustomer():
 
 
 def _getValueFromOAuth(field, credentials=None):
+    if GC_Values[GC_ENABLE_DASA]:
+        if field == 'email':
+            return GC_Values[GC_DASA_ADMIN]
```
jay0lee, line 8516: OK, I see now. Let's not do this. It won't work in most cases anyway with DASA authentication alone. For those commands that end up calling _getValueFromOAuth, we may need to do something else to work with the DASA credentials or, in the case of something like Gmail commands where a service account just won't be able to use Gmail API, we should just fail with an error that the command doesn't support DASA.

jay0lee:
Alright, looking at where we are using _getValueFromOAuth('email') some more, I think we should create a new function, something like get_admin_email() which will first check for GA_ADMIN_EMAIL env variable / config setting. GA_ADMIN_EMAIL should be set to the G Suite admin user (not the dasa email address). If GA_ADMIN_EMAIL is not set and dasa is enabled, we should error out with a message. If dasa is not enabled, we go to _getValueFromOAuth.

PR #1254, opened 2020-09-26
Add view all|company|personal to print devices

taers232c:
Update documentation


src/gam/gapi/cloudidentity/devices.py
```diff
@@ -129,11 +129,26 @@ def print_():
         if myarg in ['filter', 'query']:
           device_filter = sys.argv[i+1]
           i += 2
+        elif myarg == 'view':
```
jay0lee, line 132: This just seems like another way of doing the same thing nocompanydevices/nopersonaldevices does. What's reason for adding it?

taers232c:
Jay,

Given that only two fllters work with both company and personal devices, it seems easier to say:
gam print devices view personal filter "model: iPhone 6"
than
gam print devices nocompanydevices filter "model:iPhone 6"

The following will work even though the filter only applies to personal devices:
gam print devices view all filter "model:iPhone 6"

For common filters, e.g. type, the following gets all of the devices with one API call rather that two.
gam print devices view all filter "type:IOS"

Ross
--
ross.scroggs@gmail.com

PR #1265, opened 2020-10-09
Add a GAM command to invite a person to a course 

### LINKED ISSUE #1264: Add a GAM command to invite a person to a course
Please, add a GAM command to invite a person (as a student, teacher, or owner) to a course, so the person has the option to accept or decline the invite to the course.

ricardoseriani:
Add a GAM command to invite a person (as a student, teacher, or owner) to a course, so the person has the option to accept or decline the invite to the course.

Command description:
`gam course <CourseID> invite teachers|students|owners <UserItem>`

Closes #1264

### DIFF
--- a/src/GamCommands.txt
+++ b/src/GamCommands.txt
@@ -1249,6 +1249,7 @@ gam course <CourseID> add alias <CourseAlias>
 gam course <CourseID> delete alias <CourseAlias>
 gam course <CourseID> add teachers|students <UserItem>
 gam course <CourseID> delete|remove teachers|students <UserItem>
+gam course <CourseID> invite teachers|students|owners <UserItem>
 gam course <CourseID> sync teachers|students <UserTypeEntity>
 gam print course-participants [todrive] (course|class <CourseID>)*|([teacher <UserItem>] [student <UserItem>] [states <CourseStateList>]) [show all|students|teachers]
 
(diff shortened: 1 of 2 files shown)


jay0lee:
Thanks for the PR! Couple questions we need to answer here:
- we need methods to list invitations as well as uninvite.
- what happens if an admin tries to directly add a student who has already been invited but not accepted? Makes above more important so admin can diagnose issue but would also be good to understand behavior.

ricardoseriani:
>     * we need methods to list invitations as well as uninvite.
Good point.
Any recommendations on the syntax of the command?



>     * what happens if an admin tries to directly add a student who has already been invited but not accepted? Makes above more important so admin can diagnose issue but would also be good to understand behavior.

AFAIK, currently, if a person is invited as a student, and after that is added as a student, the person is 'automatticaly accepted' as a student, (an `add` has priority over an `invite`).

taers232c:
Here's what Advanced GAM offers
* https://github.com/taers232c/GAMADV-XTD3/wiki/Classroom-Invitations

ricardoseriani:
So, I should create a separate command for invitations, not use it as a subcommand for course?

And about uninvite, the only way that I found is by getting the `<invitation-id>`, but in that  case, I have to do 2 or more API requests, is that OK ?

jay0lee:
closing this for now but feel free to re-open a PR with recommended fixes.

PR #1269, opened 2020-10-11
Add member <UserItem> to print cigroups|cigroup-members

taers232c:
Drop Google-Coordinate product ID
Documentation cleanup

### DIFF
--- a/src/gam/gapi/__init__.py
+++ b/src/gam/gapi/__init__.py
@@ -218,7 +218,7 @@ def got_total_items_first_last_msg(items):
     return f'Got {TOTAL_ITEMS_MARKER} {items}: {FIRST_ITEM_MARKER} - {LAST_ITEM_MARKER}' + '\n'
 
 
-def process_all_pages_result(page, items, all_items, total_items, page_message, message_attribute):
+def process_page(page, items, all_items, total_items, page_message, message_attribute):
     """Process one page of a Google service function response.
 
   Append a list of items to the aggregate list of items
@@ -267,7 +267,8 @@ def process_all_pages_result(page, items, all_items, total_items, page_message, 
         sys.stderr.write(show_message)
     return (page_token, total_items)
 
-def finalize_all_pages_result(page_message):
+def finalize_page_message(page_message):
+    """ Issue final page_message """
     if page_message and (page_message[-1] != '\n'):
         sys.stderr.write('\r\n')
         sys.stderr.flush()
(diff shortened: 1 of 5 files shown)


src/gam/gapi/__init__.py
```diff
@@ -218,6 +218,60 @@ def got_total_items_first_last_msg(items):
     return f'Got {TOTAL_ITEMS_MARKER} {items}: {FIRST_ITEM_MARKER} - {LAST_ITEM_MARKER}' + '\n'
 
 
+def process_all_pages_result(page, items, all_items, total_items, page_message, message_attribute):
```
jay0lee, line 221: what are we gaining by pulling this into a function? Also the name vs description is confusing, are we processing all pages or one page here?

src/gam/gapi/cloudidentity/groups.py
```diff
@@ -222,16 +229,30 @@ def print_():
                 display.add_titles_to_csv_file([
                     'Owners',
                 ], titles)
-    gam.printGettingAllItems('Groups', None)
+    gam.printGettingAllItems('Groups', usemember)
     page_message = gapi.got_total_items_first_last_msg('Groups')
-    entityList = gapi.get_all_pages(ci.groups(),
-                                    'list',
-                                    'groups',
+    if usemember:
```
jay0lee, line 234: the challenge here is searchTransitiveGroups returns GroupRelation:

https://cloud.google.com/identity/docs/reference/rest/v1beta1/groups.memberships/searchTransitiveGroups#response-body

while groups.list just returns Groups:
https://cloud.google.com/identity/docs/reference/rest/v1beta1/groups/list#response-body

these two have some differences today but if that difference grows we end up with a whole bunch of if/then/else logic on this and it's just messy. searchTransitiveMemberships is similar.

I think it'd be cleaner to break these transitive API calls out into their own funciton (and thus own GAM command). The functions will be cleaner and easier to read and it means admins need to call another GAM command to use the transitive functions. Considering that the transitive functions only work on the Enterprise SKUs this makes sense also that some GAM commands won't work (as opposed to some arguments on some GAM commands not working).

taers232c:
I put the usemember code back
It's purpose is to select groups to print, none of the data it returns is printed.
It's use is just like print groups
```
gam print cigroups member user@domain.com
gam print groups member user@domain.com
```

taers232c:
Will re-issue for 5.24

PR #1272, opened 2020-10-15
Multiple updates

taers232c:
Add member <UserItem> to print cigroups|cigroup-members to select groups to display
Drop Google-Coordinate product ID
Update print|show driveactivity to Drive Activity API v2
Check for more parents than 1 in create|update drivefile
Update documentation

### DIFF
--- a/src/gam/__init__.py
+++ b/src/gam/__init__.py
@@ -1091,7 +1091,7 @@ def buildAlertCenterGAPIObject(user):
 
 def buildActivityGAPIObject(user):
     userEmail = convertUIDtoEmailAddress(user)
-    return (userEmail, buildGAPIServiceObject('appsactivity', userEmail))
+    return (userEmail, buildGAPIServiceObject('driveactivity', userEmail))
 
 
 def buildDriveGAPIObject(user):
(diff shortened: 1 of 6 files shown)


src/gam/__init__.py
```diff
@@ -3568,6 +3637,14 @@ def getDriveFileAttribute(i, body, parameters, myarg, update=False):
     return i
 
 
+def check_num_parents(user, body):
```
jay0lee, line 3640: function name should match what we are returning so we can do something like:

```
if has_multiple_parents(body):
    # warn user
```
and we get an idea of what has_multiple_parents does and returns just by it's name.

Also, let's move the stderr message out of the function and just make the function return true/false. We can output something within the if statement if needed.

src/gam/gapi/directory/cros.py
```diff
@@ -15,11 +16,20 @@
...
 
 
+def _display_cros_command_result(cd, device_id, command_id, times_to_check_status):
+    for _ in range(0, times_to_check_status):
+        time.sleep(2)
+        result = gapi.call(cd.customer().devices().chromeos().commands(), 'get',
+            customerId=GC_Values[GC_CUSTOMER_ID], deviceId=device_id,
+            commandId=command_id)
+        display.print_json(result)
+        if result.get('state') in {'EXPIRED', 'CANCELLED', 'EXECUTED_BY_CLIENT'}:
```
jay0lee, line 27:
```
final_states = ['EXPIRED', 'CANCELLED', 'EXECUTED_BY_CLIENT']
if result.get('state') in final_states:
....
```
it's easier to understand what the code is doing here if we declare what we believe are final states ahead of the if statement.

src/gam/gapi/directory/cros.py
```diff
@@ -40,7 +50,7 @@ def issue_command():
              body['commandType'] = command_map[command]
              i += 2
              if command == 'setvolume':
-                 body['payload'] = {'volume': int(sys.argv[i])}
+                 body['payload'] = json.dumps({'volume': sys.argv[i]})
```
jay0lee, line 54: you may need to sync to master where this change already exists.

taers232c:
Will re-issue

PR #1290, opened 2020-12-07
Add info deviceuserstate

taers232c:


src/gam/gapi/cloudidentity/devices.py
```diff
@@ -209,10 +234,9 @@ def update_state():
             i += 2
         else:
             controlflow.invalid_argument_exit(sys.argv[i], 'gam update deviceuserstate')
-    name = f'{deviceuser}/clientStates/{client_id}'
-    updateMask = ','.join(body.keys())
     result = gapi.call(ci.devices().deviceUsers().clientStates(), 'patch',
-     name=name, customer=f'customers/{customer}', updateMask=updateMask, body=body)
+                       name=f'{deviceuser}/clientStates/{client_id}', customer=customer,
```
jay0lee, line 238: Leave the variable creation as it was, it makes it easier to read and it avoids hitting line limits. Same thing for info_state above.

jay0lee: Let's also leave updateMask as is, I know behavior is same today without it but I think it's better to be explicit in setting it and I believe other API calls that use updateMask require it to be set.

PR #1294, opened 2020-12-11
Handle updating individual annotated fields

taers232c:


src/gam/gapi/cbcm.py
```diff
@@ -186,7 +186,12 @@ def update():
         else:
             controlflow.invalid_argument_exit(sys.argv[i],
                                               'gam update browser')
+    browser = gapi.call(cbcm.chromebrowsers(), 'get', deviceId=device_id,
+                        customer=GC_Values[GC_CUSTOMER_ID],
+                        projection='BASIC',
+                        fields='deviceId,annotatedAssetId,annotatedLocation,annotatedNotes,annotatedUser')
```
jay0lee, line 192: set fields as a variable before the get API call.

jay0lee: You can even do:
```
fields = list(attributes.values())
```
I don't think we actually need to get the deviceId for this.

PR #1329, opened 2021-03-12
UserInvitations clean up

taers232c:
This version of Cloud Identity API wants C in customer
I commented it out in case the developers figure out that it's inconsistent with devices and groups

Delete extraneous code


src/gam/gapi/cloudidentity/userinvitations.py
```diff
@@ -14,8 +14,8 @@
 def _get_customerid():
     ''' returns customer in format needed for this API'''
     customer = GC_Values[GC_CUSTOMER_ID]
-    if customer.startswith('C'):
-        customer = customer[1:]
+#    if customer.startswith('C'):
+#        customer = customer[1:]
```
jay0lee, line 18: just take these out. Think I was testing something at one point and missed removing them because customer normally = my_customer.

PR #1330, opened 2021-03-13
Clean up printer commands/documentation

taers232c:
driverless has to take value so it can be changed from true to false
Drop separate deleteprinters command, merge into delete printers

### DIFF
--- a/src/gam/gapi/directory/printers.py
+++ b/src/gam/gapi/directory/printers.py
@@ -1,10 +1,11 @@
 '''Commands to manage directory printers.'''
 # pylint: disable=unused-wildcard-import wildcard-import
 
+import sys
+
 from gam import controlflow
 from gam import display
 from gam import gapi
-from gam import getBoolean
 from gam import getUsersToModify
 from gam.var import *
 from gam.gapi import directory as gapi_directory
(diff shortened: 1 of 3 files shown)


jay0lee:
Driverless is set to false when makeandmodel is set. Setting driverless to false and not setting a makeandmodel is an error. I believe the API "just works" when creating/updating these two but feel free to prove me wrong with an example.

src/gam/gapi/directory/printers.py
```diff
@@ -76,15 +58,32 @@ def create():
...
     customer_id = _get_customerid()
     printer_id = sys.argv[3]
-    name = f'{customer_id}/chrome/printers/{printer_id}'
-    gapi.call(cdapi.customers().chrome().printers(),
-                       'delete',
-                       name=name)
-    print(f'Deleted printer {printer_id}')
+    if printer_id.lower() not in {'crosfile', 'croscsvfile'}:
+        name = f'{customer_id}/chrome/printers/{printer_id}'
+        gapi.call(cdapi.customers().chrome().printers(),
```
jay0lee, line 67: this is good but I don't think there's any harm in "batch deleting" 1 printer. Let's just let a list of one printer id fall through to the batch delete call.

src/gam/gapi/directory/printers.py
```diff
@@ -76,15 +58,32 @@ def create():
 
 
 def delete():
```
jay0lee, line 60: We could also use a helper function to convert user supplied displayName for printer to the printer ID. Feel free to take a crack at it. I'm pretty sure displayName does not have to be unique for printers though so if we get more than one match we'll need to error out with a list of IDs that match and whatever info we know about them so admin can be more specific.

PR #1333, opened 2021-03-13
Customer cleanup

taers232c:


src/gam/gapi/directory/customer.py
```diff
@@ -26,7 +25,7 @@ def doGetCustomerInfo():
         result = gapi.call(
             cd.domains(),
             'get',
-            customer=customer_id,
+            customer=customer_info['id'],
```
jay0lee, line 28: _get_customerid() is meant to handle CUSTOMER_ID env variable with or without C. Why undo this?

PR #1337, opened 2021-03-18
Add info printer command/ChromePolicy cleanup

taers232c:


src/gam/gapi/chromepolicy.py
```diff
@@ -31,17 +31,15 @@ def build():
...
 
 
-def print_policies():
+def printshow_policies():
     svc = build()
     customer = _get_customerid()
-    orgunit = '/'
-    printer_id = None
-    app_id = None
+    orgunit = printer_id = app_id = None
```
jay0lee, line 37: Let's keep these broken out into separate lines and assignments to avoid any future issues. What I want to avoid is something like:
```
# if value is immutable (none, int, str, etc) we're fine
a = b = c = None
a = 1
print(a, b, c)
<Outputs 1, None, None as expected>

# but if value is mutable the behavior is less obvious
a = b = c = []
a.append(1)
print(a, b, c)
<outputs [1] [1] [1] - all variables point to the same value>
```
we sometimes change these initial values at a later time (like when we decide orgunit should be none). So let's standardize on not doing this one line assignments to avoid any such issues in the future.

PR #1340, opened 2021-03-22
Match ENUM fields and descriptions

taers232c:
```
    for schema in schemas:
        schema_name = schema.get('name', '').split('/')[-1]
        schema_dict = {
                'name': schema_name,
                'description': schema.get('policyDescription', ''),
                'settings': {},
                }
        field_descriptions = schema.get('fieldDescriptions', [])
        for mtype in schema.get('definition', {}).get('messageType', {}):
            for setting in mtype.get('field', {}):
                setting_name = setting.get('name', '')
```
There's lots of .get() in here; is it really the case that a schema might not have a name or a setting might not have a name or a schema has no fieldDescriptions?

### DIFF
--- a/src/gam/gapi/chromepolicy.py
+++ b/src/gam/gapi/chromepolicy.py
@@ -111,6 +111,7 @@ def build_schemas(svc=None, sfilter=None):
                 'description': schema.get('policyDescription', ''),
                 'settings': {},
                 }
+        field_descriptions = schema.get('fieldDescriptions', [])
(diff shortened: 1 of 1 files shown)


jay0lee:
In some cases yes, in others I'd prefer we play it safe. New policies will be coming online all the time and Google does not state which fields will always be present. It's annoying to need to do an emergency GAM release because Google added a new policy that didn't have the normal fields.

Issue #1345, opened 2021-03-25
Add argument to prevent creating user/group/alias addresses that would evict an unmanaged account.


jay0lee:
We've added support for the User Invitation API. However it'd be good if an admin could say something like:
```
gam create user <email> ... ifnotinvitable
```
that would tell GAM to check if the new user email address is invitable (meaning it's currently an unmanaged account) and refuse to create the address if it is. This allows admins to avoid accidentally evicting unmanaged accounts.

Ideally we'd do this on user create, user update if email addresses change, group create/update and alias create/update.

PR #1352, opened 2021-04-02
Implement Issue #1345

taers232c:


src/gam/__init__.py
```diff
@@ -8262,6 +8280,10 @@ def doUpdateUser(users, i):
                 'primary': False,
                 'address': user_primary
             }]
+        if (verifyNotInvitable and'primaryEmail' in body and
+            gapi_cloudidentity_userinvitations.get_is_invitable_user(body['primaryEmail'])):
+            sys.stderr.write(f'User {user} not updated, new primaryEmail {body["primaryEmail"]} is an unmanaged account\n')
```
jay0lee, line 8285: should be calling controlflow.system_error_exit() here instead of sys.exit directly.

src/gam/__init__.py
```diff
@@ -7020,6 +7020,9 @@ def _splitSchemaNameDotFieldName(sn_fn, fnRequired=True):
             else:
                 body[up][schemaName][fieldName] = sys.argv[i]
             i += 1
+        elif myarg == 'verifynotinvitable':
+            body['verifyNotInvitable'] = True
```
jay0lee, line 7024: It's pretty ugly that were over-extending body here. I'd prefer to see us return a tuple of verify_not_invitable and body.

PR #1354, opened 2021-04-06
Implement Chrome version history

taers232c:


src/gam/__init__.py
```diff
@@ -995,6 +998,12 @@ def buildGAPIObject(api):
     return service
 
 
+def buildGAPIObjectNoAuthentication(api):
+    GM_Globals[GM_CURRENT_API_USER] = None
+    http = transport.create_http(cache=GM_Globals[GM_CACHE_DIR])
```
jay0lee, line 1003: httpc or http_client to avoid conflict with the Python http module.

taers232c: I don't understand

taers232c: Oh, now I do

PR #1362, opened 2021-04-08
Restandardize chromehistory columns; fix chromepolicy

taers232c:


src/gam/gapi/chromehistory.py
```diff
@@ -84,14 +84,16 @@ def _get_platform_map(cv):
...
-        key = platform.replace('_', '')
-        platform_map[key] = platform
+    for cplatform in platforms:
+        key = cplatform.replace('_', '')
+        platform_map[key] = cplatform
     return platform_map
 
 
-def _get_channel_map(cv):
+def get_channel_map(cv=None):
```
jay0lee, line 93: I don't have a strong preference whether these are private or not but we should keep get_channel_map and get_platform_map the same.

PR #1372, opened 2021-04-23
Googleapiclient test fix

jay0lee:


jay0lee:
@ejochman googleapiclient >= 2.1 broke tests. Can you review this fix?

ejochman:
At the time, it wasn't clear in the `HttpError` docs as to what type was expected as the `resp` parameter. But, if things changed in `googleapiclient` such that it now requires a proper `Response` object, then this looks good to me.

PR #1378, opened 2021-05-04
Cloud Identity v1 only uses preferredMemberKey

taers232c:


src/gam/__init__.py
```diff
@@ -9018,32 +9018,32 @@ def user_lic_result(request_id, response, exception):
...
+                group_email = group_name_mapping[group_name]
+                for edge in adj.get('edges', []):
+                    seen_group_count[group_email] = seen_group_count.get(group_email, 0) + 1
+                    member_email = edge.get('preferredMemberKey', {}).get('id')
+                    edges.append((member_email, group_email))
+            print_group_map(user_email, group_displayname_mapping, seen_group_count, edges, 3, 'direct')
+            if seen_group_count and max(seen_group_count.values()) > 1:
+                print()
+                print('   * user has multiple direct or inherited memberships in group')
             print()
```
jay0lee, line 9046: move the last print() out of the if statement so that always a blank line between Group Map and licenses.

taers232c: It's already that way:
```
    elif getCIGroups:
        memberships = gapi_cloudidentity_groups.get_membership_graph(user_email)
        if memberships:
            print('\nGroup Membership Tree:')
           ...
            if seen_group_count and max(seen_group_count.values()) > 1:
                print()
                print('   * user has multiple direct or inherited memberships in group')
            print()
 ```
The initial print('\nGroup Membership Tree:') and final print() are at the same level; neither appear if there are no memberships.

**Comment by jay0lee** [comment]
Let's keep using the beta1 API for those functions that access memberKey.

File: src/gam/__init__.py (hunk 1 above)
Comment (jay0lee) [reply src/gam/__init__.py:9046]: So we should pull them both out then so that we end up printing something like:

```
Group Member Tree:

Licenses:
....
```
which indicates there are no memberships. Saying nothing about the tree might seem like the argument just didn't work.

PR #1379, opened 2021-05-06
Code cleanup; display role for group members

taers232c:


src/gam/gapi/cloudidentity/groups.py
```diff
@@ -118,43 +118,40 @@ def info():
...
-        if member_id.isdigit():
-            print(f'{" " * spaces}{member_email} - user')
-        else:
-            print(f'{" " * spaces}{member_email} - group')
-            group_id = group_email_to_id(ci, member_email)
-            print_member_tree(ci, group_id, spaces + 2)
+        role = get_single_role(member.get('roles', [])).lower()
+        email = member.get('memberKey', {}).get('id')
+        member_type = member.get('type', 'USER').lower()
+        print(f'{" " * spaces}{role}: {email} ({member_type})')
```
jay0lee, line 152: groups can only be members and thus all users that are inherited are members so I suggest we just print "member" on the group and not all the way down the tree.

PR #1393, opened 2021-07-02
Fix page_args_in_body, update namespace handling in show chromepolicies

taers232c:
When page_args_in_body is true you have to add body to kwargs to ensure a place for pageToken

Allow setting a list of namespaces that override the defaults for printerid (not likely) and appid.

### DIFF
--- a/src/gam/gapi/__init__.py
+++ b/src/gam/gapi/__init__.py
@@ -319,11 +319,12 @@ def get_all_pages(service,
   Returns:
     A list of all items received from all paged responses.
   """
+    if page_args_in_body:
+        kwargs.setdefault('body', {})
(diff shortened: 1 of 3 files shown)


src/gam/gapi/__init__.py
```diff
@@ -319,11 +319,12 @@ def get_all_pages(service,
   Returns:
     A list of all items received from all paged responses.
   """
+    if page_args_in_body:
+        kwargs.setdefault('body', {})
```
jay0lee, line 323: Does the same thing in two lines that my code did in 1?

taers232c: This is to ensure that body gets set in kwargs in case _get_max_page_size_for_api_call returns None. Notice that _get_max_page_size_for_api_call always returns {'maxResults': max_results} when it should return {'pageSize': max_results} for Chrome Policies.

PR #1417, opened 2021-08-26
Make GAM pip-installable

janosh:
This is just a rough draft to get the ball rolling on making GAM `pip`-installable and closing #720. I hope to get some feedback since GAM has a slightly unusual package structure and I'm unfamiliar with GAMs current installation procedure.

### DIFF
--- a/src/setup.cfg
+++ b/src/setup.cfg
@@ -1,5 +1,5 @@
 [metadata]
-name = Google Apps Manager
+name = GAM for Google Workspace
 version = 6.0.7
 description = Command line management for Google Workspaces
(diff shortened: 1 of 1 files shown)


janosh:
> No, but I suspect if you remove src/ from the license line in setup.cfg it will solve it.

Yes, I was going to submit a PR for that but awaiting your response on whether to mention pip install in readme.

> Does pip really use .rtf format? The LICENSE file might be a battery choice there for plaintext.

Not sure. Support in what sense? It doesn't actually display it anyway, does it?

Are you planning to release GAM on PyPI?

PR #1419, opened 2021-08-28
Fix pip license error + add pip install command to readme

### LINKED ISSUE #720: Make GAM pip installable

janosh:
As a follow-up to #1417, this should fix the [pip license error](https://github.com/jay0lee/GAM/pull/1417#issuecomment-907418598) and inform users through the readme that GAM can now be installed with `pip`.

Closes #720.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,23 +1,46 @@
 GAM is a command line tool for Google Workspace (fka G Suite) Administrators to manage domain and user settings quickly and easily.
 
 ![Build Status](https://github.com/jay0lee/GAM/workflows/Build%20and%20test%20GAM/badge.svg)
+
 # Quick Start
+
 ## Linux / MacOS
+
 Open a terminal and run:
-```
+
+```sh
 bash <(curl -s -S -L https://git.io/install-gam)
 ```
+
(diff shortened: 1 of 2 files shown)


taers232c:
I got this: warning: the 'license_file' option is deprecated, use 'license_files' instead

jay0lee:
Let's hold off on the readme changes until this is stable and (maybe) on
PyPi.

taers232c:
Where do the discovery document .json files reside?
Where does GamCommands.txt reside?

janosh:
> Let's hold off on the readme changes until this is stable and (maybe) on

@jay0lee Sure.

> I got this: warning: the 'license_file' option is deprecated, use 'license_files' instead

Interesting. [This deprecation is somewhat recent](https://github.com/pypa/setuptools/pull/2620) but I'm still surprised I've never seen this warning before.

PR #1426, opened 2021-09-17
Several fixes/updates

taers232c:
agreedToTerms is now read-only
Fix sync devices


taers232c:
Jay,

Note that the Cloud Identity Devices API is not returning the assetTag value; specifying it in sync will be problematic.

Ross

PR #1459, opened 2021-12-09
languages update; fields for gam info user; cloud identity groups update to v1

taers232c:
The API doesn't return languages unless you specifically mention in in fields list

### DIFF
--- a/src/gam/__init__.py
+++ b/src/gam/__init__.py
@@ -8804,6 +8804,7 @@ def doGetUserInfo(user_email=None):
             i = 4
         else:
             user_email = _get_admin_email()
+    fieldsList = []
     getSchemas = True
     getAliases = True
     getGroups = True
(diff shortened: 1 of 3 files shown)


src/gam/__init__.py
```diff
@@ -9814,6 +9840,9 @@ def doPrintUsers():
                     if user_email.find('@') != -1:
                         user['primaryEmailLocal'], user[
                             'primaryEmailDomain'] = splitEmailAddress(user_email)
+                up = 'languages'
+                if up in user:
```
jay0lee, line 9844: No reason to keep up here.

src/gam/__init__.py
```diff
@@ -6703,6 +6703,7 @@ def _splitSchemaNameDotFieldName(sn_fn, fnRequired=True):
                 if language.lower() in LANGUAGE_CODES_MAP:
                     lang_item['languageCode'] = LANGUAGE_CODES_MAP[language.lower()]
                 else:
+                    lang_item.pop('preference', None)
```
jay0lee, line 6706: this should not be necessary as preference is not set if customLanguage is set:

https://github.com/jay0lee/google-api-tracker/blob/master/admin-directory_v1.json#L7367

taers232c:
The user can say: gam update user billybob languages drawl+,en-US-
preference is set from +- before the language is detected as custom
so we clear it if set on a custom language
--
***@***.***

jay0lee:
I'd prefer to see that throw an API error (which the user can then correct) as opposed to making assumptions about what the user meant to enter. Suppose user mis-typed a non-customer language?

PR #1464, opened 2021-12-29
Make gam info crostelemetry <SerialNumber>

taers232c:
I apologize for missing when you added crostelemetry, otherwise I'd be making these changes before you released.

### DIFF
--- a/src/gam/gapi/chromemanagement.py
+++ b/src/gam/gapi/chromemanagement.py
@@ -203,7 +203,7 @@ def printAppDevices():
     display.write_csv_file(csvRows, titles, 'Chrome Installed Application Devices', todrive)
 
 
-def printShowCrosTelemetry(show=False):
+def printShowCrosTelemetry(mode):
     cm = build()
     cd = None
     parent = _get_customerid()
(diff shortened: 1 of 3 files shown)


src/gam/gapi/chromemanagement.py
```diff
@@ -217,6 +217,10 @@ def printShowCrosTelemetry(show=False):
     supported_readmask_values.sort()
     supported_readmask_map = {item.lower():item for item in supported_readmask_values}
     i = 3
+    if mode == 'info':
+        filter_ = f'serialNumber={sys.argv[i]}'
```
jay0lee, line 221: Need to handle user running just `gam info crostelemetry` with no serial.

taers232c: All gam info commands take an object identifier, use gam show to do a broader selection

jay0lee: My point is you are doing `sys.argv[i]` (where i == 3) without confirming `len(sys.argv) >= 4`. If the user runs simply `gam info crostelemetry` it only get's caught as an IndexError way up in __init__.py and the user gets a useless generic message. We should confirm user entered a serial number and throw a (nicer) error if not.

I'm sure we don't catch the same thing in other info commands but that doesn't mean we should make the same mistakes here.

PR #1472, opened 2022-01-15
Improve gam user show vaultholds

taers232c:


src/gam/gapi/vault.py
```diff
@@ -668,26 +669,34 @@ def updateHold():
...
+                    if orgUnitId in orgUnits:
+                        print(f'FOUND: OrgUnit {orgUnits[orgUnitId]} for user {user} is on hold in matterId {matterId} and holdId {hold["holdId"]} named "{hold["name"]}"')
+                        totalHolds += 1
+                else:
+                    for account in hold.get('accounts', []):
+                        if user == account.get('email', '').lower():
+                            print(f'FOUND: User account {user} is on hold in matterId {matterId} and holdId {hold["holdId"]} named "{hold["name"]}"')
+                            totalHolds += 1
+                            break
+    sys.exit(totalHolds)
```
jay0lee, line 699: If total holds > 255 this will return a bad result. We also shouldn't be exiting with non-zero error code on success.

PR #1476, opened 2022-02-05
Document create admin condition

taers232c:


taers232c:
I've got another PR to handle the update alias issue, do you want to accept/reject this one first?

PR #1481, opened 2022-02-12
Add verifytarget to update alias to address Issue #1479

taers232c:


jay0lee:
I get what you did with all the try and throw reasons but how often do you think GAM does all that just to keep the API call from causing Python to throw an error? I wonder if it makes sense to have something like.a return_nothing_on_errors passed into gapi.call? If the given errors triggered we'd just return something like {}.

Otherwise looks good but just.a Friday evening thought to save us some trouble

src/gam/__init__.py
```diff
@@ -8327,22 +8327,48 @@ def doRemoveUsersAliases(users):
...
     target_email = normalizeEmailAddressOrUID(sys.argv[5])
-    if len(sys.argv) > 6:
-        myarg = sys.argv[6].lower().replace('_', '')
-        if myarg != 'verifynotinvitable':
+    i = 6
+    while i < len(sys.argv):
+        myarg = sys.argv[i].lower().replace('_', '')
+        if myarg == 'verifynotinvitable':
+            if gapi_cloudidentity_userinvitations.is_invitable_user(alias):
+                controlflow.system_error_exit(51, f'Alias not updated, {alias} is an unmanaged account')
```
jay0lee, line 8360: Doesn't this mean it's not an alias at all? It can't be both an alias and an unmanaged account.

src/gam/__init__.py
```diff
@@ -8327,22 +8327,48 @@ def doRemoveUsersAliases(users):
...
+            i += 1
+        elif myarg == 'verifytarget':
```
jay0lee, line 8362: Suppose we just made this the default behavior? How many API calls would that add to the command (worst and best case scenarios)

taers232c: If you specify user|group, there is one extra call.
If you specify target then there is one extra call if the target is a user, two if the target is a group, two if the target doesn't exist

src/gam/__init__.py
taers232c: verifynotinvitable is verifying that the alias is not an unmanaged account; should it be verifying that the target is not an unmanaged account.

jay0lee:
Got it. My vote is to just make it the default behavior and maybe add an
option to NOT check if admin is certain the target already exists.

src/gam/__init__.py
jay0lee: If th alias already exists and we find it then it's not an unmanaged account.

taers232c: verifyTarget is verifying that the **target** exists.
verifyNotInvitable is verifying that the **alias** is not an unmanaged account.
Aren't these two checks independent or should verifyNotInvitable be checking that we're not trying to make an alias for an unmanaged account or should it be checking that neither address is an unmanaged account.

Isn't there some magic about making an alias to an unmanaged account and then deleting the alias?

jay0lee:
The unmanaged account can only exist if the alias address does not exist (you can't have an unmanaged account for any user, group or alias address in Workspace). So the unmanaged check is unnecessary, we'd fail on deleting the alias anyway.

taers232c:
Should the verifynotinvitable code be taken out of create alias as well? They're both checking the alias about to be created.

taers232c:
Sorry for being slow here, but in both cases, the verifynotinvitable check is against the alias we are creating, not the target.

Additionally, I'm curious about the invitable check itself
```
$ gam print userinvitations
Getting all User Invitations in Google Workspace account (may take some time on a large account)...
Got 3 User Invitations...
name,state,updateTime,mailsSentCount
conflict2@clouddevices.us,ACCEPTED,2018-04-30T14:05:24.902Z,1
consumer-ross@s.jaylee.us,NOT_YET_SENT,2021-03-15T21:17:10.381Z,2
consumer-test1@s.jaylee.us,NOT_YET_SENT,2021-03-12T16:50:14.365Z,
$ gam check userinvitation consumer-ross@s.jaylee.us
 isInvitableUser: False
$ gam check userinvitation conflict2@clouddevices.us
 isInvitableUser: False
```

jay0lee:
But update alias is not creating a new address, it's moving the alias to a
new user/group. If both deletes fail then we know it's not a managed
address and we fail on the update there.

The list invites command pulls an indexed list of uanaged users whereas the
check invitable command does a "real time" check on the address. It can
take 24-48 hours for an address that was evicted, accepted invite or
renamed out of the domain on their own to appear in the list so the check
is more accurate to current status while the list is more complete (but
lags)

PR #1482, opened 2022-02-14
Update LICENSE file contents and location

ejochman:
Move LICENSE to root and remove non-applicable "server subcomponents"
section.


jay0lee:
What's the reason for moving it to root? That will need to be updated in other build scripts also.

ejochman:
The root is a conventional location for license files in github projects. There are some automations that expect the license to be in that location. I've symlink'd `src/LICENSE` (the old location) to the new root location so that build scripts should continue to work as expected (AFAIK).

PR #1486, opened 2022-02-24
Cleanup

taers232c:


src/gam/auth/oauth.py
```diff
@@ -605,15 +604,14 @@ def run_dual(self,
...
-        print(f'URL is:     {self.auth_url}')
+        print(f"URL is:     {d['auth_url']}")
         user_input.start()
         while True:
             sleep(0.1)
             if not http_client.is_alive():
                 user_input.terminate()
                 break
-            elif not user_input.is_alive():
+            if not user_input.is_alive():
```
jay0lee, line 614: We don't want to do this. If http_client finishes first we'll terminate user_input which will then cause this line to return true  and then we end up trying to terminate http_client which already completed successfully.

Only one terminate() should ever be necessary to kill the process that did not complete.

taers232c: We'll never get to the line because after terminating user_input we break out of the loop; only one termination is done

dldinternet: Some thoughts ... 
- what if True == not (http_client.is_alive() or user_input.is_alive())
- The suggested change has no effect UNLESS user_input.terminate() does not result in  not user_input.is_alive() == True
Truth table before the change:
 `|------------------------+------------------------+------------------------|------------------------|------------------------|
| http_client.is_alive() | user_input.is_alive()) | http_client.terminate()| user_input.terminate() |    break               |
|------------------------+------------------------+------------------------+------------------------+------------------------|
|          yes           |         yes            |          no            |           no           |           no           |
|          yes           |         no             |          yes           |           no           |           yes          |
|          no            |         yes            |          no            |           yes          |           yes          |
|          no            |         no             |          no            |           yes          |           yes          |
|------------------------+------------------------+------------------------+------------------------+------------------------|`
Truth table AFTER the change:
`|------------------------+------------------------+------------------------|------------------------|------------------------|
| http_client.is_alive() | user_input.is_alive()) | http_client.terminate()| user_input.terminate() |    break               |
|------------------------+------------------------+------------------------+------------------------+------------------------|
|          yes           |         yes            |          no            |           no           |           no           |
|          yes           |         no             |          yes           |           no           |           yes          |
|          no            |         yes            |          no            |           yes          |           yes          |
|          no            |         no             |          no            |           yes          |           yes          |
|------------------------+------------------------+------------------------+------------------------+------------------------|

`

PR #1520, opened 2022-05-02
Handle missing type field in multivalued schema value

taers232c:


dupuy26 (approved):

taers232c:
Jay,

Should we simply set type to work on input if no type is specified?

Ross

PR #1522, opened 2022-05-05
Multivalued schema files default to type work; update URLs

taers232c:


src/gam/var.py
```diff
@@ -11,7 +11,7 @@
 GAM_VERSION = '6.21'
 GAM_LICENSE = 'Apache License 2.0 (http://www.apache.org/licenses/LICENSE-2.0)'
 
-GAM_URL = 'https://jaylee.us/gam'
```
jay0lee, line 14: this one also.

PR #1527, opened 2022-05-28
Eliminate orgUnitPath encoding

taers232c:
% no longer needs encoding and + is handled incorrectly in API with or without encoding


taers232c:
See: https://github.com/googleapis/google-api-python-client/issues/1812

PR #1554, opened 2022-08-30
Out with the oobs

taers232c:


jay0lee:
Looking at:

https://developers.google.com/identity/protocols/oauth2/native-app#exchange-authorization-code

we need to set redirect_uri to something so just replace oob with http://127.0.0.1:8080/ like in the example.

PR #1561, opened 2022-09-14
Low memory/shelve cleanup

taers232c:


src/gam/gapi/__init__.py
```diff
@@ -242,13 +242,14 @@ def process_page(page, items, all_items, total_items, page_message, message_attr
...
         total_items += num_page_items
-        if type(all_items) is list:
-            all_items.extend(page_items)
-        elif all_items is not None:
-            i = len(all_items)
-            for item in page_items:
-                all_items[str(i)] = item
-                i += 1
+        if all_items is not None:
+            if isinstance(all_items, list):
```
jay0lee, line 246: I don't see the value in multiple levels of if here.

taers232c:
Personal preference; back to the original.

PR #1578, opened 2022-11-18
Code fix, consistency preference

taers232c:


src/gam/gapi/cloudidentity/inboundsso.py
```diff
@@ -76,18 +76,17 @@ def profile_displayname_to_name(displayName, ci=None):
...
+                                  filter=_filter)
     matches = []
     for profile in profiles:
         if displayName.lower() == profile.get('displayName', '').lower():
             matches.append(profile)
     if len(matches) == 1:
         return matches[0]['name']
-    elif len(matches) == 0:
-        controlflow.system_error_exit(3, f'No Inbound SSO profile matching the name {displayName}')
+    if len(matches) == 0:
```
jay0lee, line 86: elif this so we don't need to re-eval if len(matches) == 1 above.

taers232c: If len(matches) == 1 we return; they'll be no re-evaluation

src/gam/gapi/cloudidentity/inboundsso.py
```diff
@@ -251,24 +252,19 @@ def create_credentials():
         elif myarg == 'keysize':
             key_size = int(sys.argv[i+1])
             if key_size not in allowed_sizes:
-                controlflow.expected_argument_exit('key_size',
-                                                   allowed_sizes,
-                                                   key_size)
+                controlflow.expected_argument_exit('key_size', allowed_sizes, key_size)
```
jay0lee, line 255: wrap should stay in place unless we can keep a single line under 80 char

src/gam/gapi/cloudidentity/inboundsso.py
jay0lee: fair enough

src/gam/gapi/cloudidentity/inboundsso.py
taers232c: Do you want me to unwrap the over 80 char ones?

PR #1605, opened 2023-02-24
Add gam version checkrc

taers232c:


src/gam/__init__.py
```diff
@@ -589,7 +589,7 @@ def _getCfgRowFilter(itemName):
     GM_Globals[GM_ENABLEDASA_TXT] = os.path.join(
         GC_Values[GC_CONFIG_DIR], FN_ENABLEDASA_TXT)
     if not GC_Values[GC_NO_UPDATE_CHECK]:
-        doGAMCheckForUpdates()
+        doGAMCheckForUpdates(0)
```
jay0lee, line 592: make this:
```
doGAMCheckForUpdates(forceCheck=0)
```
so that it's obvious what the 0 is doing w/o needing to look at the function def.

Also why is this an integer instead of a bool?

PR #1609, opened 2023-03-03
Allow allowcontentmanagerstosharefolders as a shared drive restriction

taers232c:


jay0lee:
This is already a valid argument assuming the discovery file has been updated making it a valid argument. If we try to force it now it won't work as expected for admins

PR #1618, opened 2023-04-05
Two updates

taers232c:
…elector>

### DIFF
--- a/src/gam/var.py
+++ b/src/gam/var.py
@@ -657,7 +657,7 @@ GOOGLEDOC_VALID_EXTENSIONS_MAP = {
         '.docx', '.html', '.odt', '.pdf', '.rtf', '.txt', '.zip'
     ],
     MIMETYPE_GA_PRESENTATION: ['.pdf', '.pptx', '.odp', '.txt'],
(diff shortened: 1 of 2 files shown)


src/gam/var.py
```diff
@@ -657,7 +657,7 @@
         '.docx', '.html', '.odt', '.pdf', '.rtf', '.txt', '.zip'
     ],
     MIMETYPE_GA_PRESENTATION: ['.pdf', '.pptx', '.odp', '.txt'],
-    MIMETYPE_GA_SPREADSHEET: ['.csv', '.ods', '.pdf', '.xlsx', '.zip'],
+    MIMETYPE_GA_SPREADSHEET: ['.csv', '.ods', '.pdf', '.tsv', '.xlsx', '.zip'],
```
jay0lee, line 660: This is fine. Worth noting though that the export file type data is available dynamically for a user with Drive's about.get() API call. That offer's the full MIME type but we can convert that to file extensions with:
```
>>> import mimetypes
>>> mimetypes.guess_extension('text/tab-separated-values')
'.tsv'
```
I'm not 100% on this but since Google is offering the export formats dynamically for a given user the choices may vary based on user license (or may vary in the future).

PR #1774, opened 2025-04-23
Fix typos

patrickhaley:


taers232c:
Patrick,

Thanks, these will be corrected in next update.

Ross

PR #1815, opened 2025-08-11
Added Alert Center commands for displaying and setting Settings #1814

rturner-edjuster:
I am by no means familiar with the Gam code, or really Python either. I had a need to set these settings, so I thought it would be good to "send the changes back" for others to also be able to use. Hopefully this is close to what you would want in terms of a PR. This seems to work in my very limited testing. I did not try "unsetting" the topic, but that might require a different command syntax to support that. Suggestions welcome of couse.


taers232c:
Commands added with other code.

PR #1868, opened 2025-12-16
docs(README): add mise alternative method installation

jylenhof:


jay0lee:
Thanks for offering this but only official install methods can be included in our docs.

